PR #1, opened 2023-05-22
ci: Add lint workflow for GitHub Actions

siteshwar:
Related: https://gitlab.cee.redhat.com/covscan/covscan/-/issues/84

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,28 @@
+name: Lint
+on: [push]
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    container: fedora:37
+    steps:
+      - name: Install packages required for lint
+        run: |
+          dnf install -y pre-commit make git-core
+      - name: Checkout Git Repository
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Lint
+        run: |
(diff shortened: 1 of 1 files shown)


isinyaaa:
I think we'd be better off using actions for this. We're basically checking out and running pre commit, and extending this should be much easier than setting linters in a custom container.

My suggestion would be to use this: https://github.com/pre-commit/action

siteshwar:
> I think we'd be better off using actions for this. We're basically checking out and running pre commit, and extending this should be much easier than setting linters in a custom container.
> 
> My suggestion would be to use this: https://github.com/pre-commit/action

This is a third party action which is currently in [maintenance mode only](https://github.com/pre-commit/action/commit/5f528da5c95691c4cf42ff76a4d10854b62cbb82) and the author seems to recommend a [separate CI service](https://pre-commit.ci). I want to keep our dependence on third party services to minimal.

Our usage of `pre-commit` is rather basic and there is no [official](https://github.com/actions) GitHub action for it. If we want to use separate action for it, we would have to maintain it on our own. This is something I would like to avoid unless there is really good reason for it.

jamacku (approved):
LGTM, works as expected.

- https://github.com/jamacku/openscanhub/actions/runs/5054414590/jobs/9069313091

PR #14, opened 2023-06-21
docs: expect local OSH instance in the ET interface examples

kdudka:
The development environment documentation was inconsistent.  It first described how to set up a local development environment but then it described how to emulate ET-initiated scans on a staging instance.

Reported-by: Juan Perez de Algaba Sierra
Closes: https://github.com/openscanhub/openscanhub/pull/14

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -129,26 +129,20 @@ As pointed above, all of these dependencies are automatically set up in the clie
 * create Errata Scan using password authentication - new pkg:
 ```sh
 osh/hub/scripts/osh-xmlrpc-client.py \
-    --hub https://$HOSTNAME/osh/xmlrpc/kerbauth/ \
-    --username=kdudka --password=xxxxxx \
-    create-scan -t curl-7.29.0-25.el7 \
-    --et-scan-id 1234 --advisory-id 4567 \
-    --owner kdudka --release RHEL-7.7.0 \
-    --base NEW_PACKAGE
+    --hub http://osh-hub:8000/xmlrpc/kerbauth/ --username=user --password=xxxxxx \
+    create-scan -b NEW_PACKAGE -t libssh2-1.10.0-7.fc38 --et-scan-id=1 \
+    --release=Fedora-37 --owner=admin --advisory-id=1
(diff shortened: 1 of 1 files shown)


siteshwar:
Here are few thoughts about this PR:

- Why are you self referencing the PR in the commit?
- We should continue to use full URLs while referencing issues. This should be documented in the contribution docs.
- We should either use `Resolves` or `Closes` consistently. We have been using `Resolves`, so I believe we should prefer that.

kdudka:
> Why are you self referencing the PR in the commit?

So that readers of the `git log` output can easily access the review comments associated with the commit.

> We should continue to use full URLs while referencing issues.

I fully agree but this was reported over Slack today.

> We should either use `Resolves` or `Closes` consistently.

I think that `Closes` is commonly used for pull requests whereas `Resolves` is used for issues.

@lzaoral has already notified me about other sloppy mistakes in this pull request.  I will update it soon.

docs/development.md
```diff
@@ -129,26 +129,20 @@ As pointed above, all of these dependencies are automatically set up in the clie
...
-    --base curl-7.29.0-25.el7
+    --hub http://osh-hub:8000/xmlrpc/kerbauth/ --username=user --password=xxxxxx \
+    create-scan -b libssh2-1.10.0-5.fc37 -t libssh2-1.10.0-7.fc38 --et-scan-id=1 \
+    --release=Fedora-37 --owner=admin --advisory-id=1
 ```
 
-* create Errata Scan using Kerberos authentication:
+* create Errata Scan using Kerberos authentication (will not work in a local development environment):
 ```sh
 osh/hub/scripts/osh-xmlrpc-client.py \
```
lzaoral, line 147: I believe that this section should be updated as well.

kdudka: As discussed in person, this section is rarely useful for external users.  Once we document how to configure Kerberos authentication for OpenScanHub off Red Hat infrastructure, we will update this section accordingly.

lzaoral (approved):

siteshwar (approved):

PR #18, opened 2023-06-22
client: add completions for ZSH 

lzaoral:
Adds completions for zsh.

### DIFF
--- a/osh.spec
+++ b/osh.spec
@@ -176,7 +176,7 @@ rm -rf %{buildroot}%{python3_sitelib}/scripts
 %attr(755,root,root) %{_bindir}/osh-cli
 %{_bindir}/covscan
 %attr(644,root,root) %config(noreplace) /etc/osh/client.conf
-%{bash_completions_dir}/osh-cli.bash
+%{bash_completions_dir}
 %{zsh_completions_dir}
 %{python3_sitelib}/osh/client
 %{python3_sitelib}/osh-*-py%{python3_version}.egg-info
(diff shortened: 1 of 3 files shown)


osh.spec
```diff
@@ -175,7 +176,10 @@ rm -rf %{buildroot}%{python3_sitelib}/scripts
 %attr(755,root,root) %{_bindir}/osh-cli
 %{_bindir}/covscan
 %attr(644,root,root) %config(noreplace) /etc/osh/client.conf
+%dir %{bash_completions_dir}
 %{bash_completions_dir}/osh-cli.bash
+%dir %{zsh_completions_dir}
+%{zsh_completions_dir}/_osh-cli
```
kdudka, line 182: As noted on the [original review](https://gitlab.cee.redhat.com/covscan/covscan/-/merge_requests/398#note_6865254), this can be simplified as:
```
%{bash_completions_dir}
%{zsh_completions_dir}
```

`rpmbuild` will pick the whole subtrees, each consisting of one directory and one regular file.

kdudka (approved):

osh.spec
lzaoral: Addressed in 77ffee2 and a97c8c5.

lzaoral:
Yes, it is related to `qpid-proton` FTI: https://bugzilla.redhat.com/show_bug.cgi?id=2019312.  There is also a bz for a new upstream version so I'm going to test it locally if it unblocks the build.

PR #19, opened 2023-06-22
client: fix downloading of results for non-RPM tasks

lzaoral:
Since 1fce0c12ad4ab5fafef7a503c462b479f1d2cc60, all tasks submitted through the client have a `result_filename` argument that defines the filename without extensions of the tarball with scan results.  So we should use if it is available.  Otherwise, fall back on the previous implementation.

Related: 1fce0c12ad4ab5fafef7a503c462b479f1d2cc60 ("support scanning of tarballs")

### DIFF
--- a/osh/client/commands/cmd_diff_build.py
+++ b/osh/client/commands/cmd_diff_build.py
@@ -184,7 +184,7 @@ is not even one in your user configuration file \
 
             # store results if user requested this
             if self.results_store_file is not None:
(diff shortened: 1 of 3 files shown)


lzaoral:
@siteshwar It looks like the CI got hit by the following Ubuntu issue: https://bugs.launchpad.net/ubuntu/+source/libpod/+bug/2024394

kdudka (approved):
No changes since the last approval on internal GitLab.

PR #25, opened 2023-06-26
Add shell linter - Differential ShellCheck

jamacku:
Differential ShellCheck is a GitHub Action that performs differential ShellCheck scans on shell scripts changed via PR and reports results directly in PR.

Your scripts are in great shape, but I think you might find the `differential-shellcheck` Action useful. It is able to produce reports in SARIF format. GitHub understands this format and is able to display it nicely as a PR comment, and on the `Files Changed` tab, please see below.

image

image

Documentation is available at [@redhat-plumbers-in-action/differential-shellcheck](https://github.com/redhat-plumbers-in-action/differential-shellcheck#readme). Let me know If you are missing some feature or setting. I'm always happy to extend functionality.

### DIFF
--- a/.github/workflows/differential-shellcheck.yml
+++ b/.github/workflows/differential-shellcheck.yml
@@ -1,0 +1,35 @@
+name: Differential ShellCheck
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    permissions:
+      security-events: write
+
+    steps:
+      - name: Repository checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - id: ShellCheck
+        name: Differential ShellCheck
+        uses: redhat-plumbers-in-action/differential-shellcheck@v4
+        with:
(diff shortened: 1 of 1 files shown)


siteshwar:
@jamacku This PR introduces 2 changes of different nature:

- You are introducing `Dependabot` in the CI, which is a good change and should be approved.
- I am skeptical about introducing Differential ShellCheck, as in the long run all the changes introduced in OpenScanHub should be scanned through OpenScanHub and ShellCheck is one of the tools we use in OpenScanHub backend. So, using differential shellcheck through GitHub Actions CI could be a temporary solution for now.

In future, please keep such changes separate to make them easier to review and merge.

jamacku:
Ok, I'll split it into two PRs. 

@siteshwar Are you interested in differential ShellCheck Action, or should I drop the commit from PR since you are planning to use OpenScanHub directly?

kdudka:
The sooner we enable source code scanners in the CI the better.  I cannot see any good reason to not enable Differential ShellCheck when it is ready to be enabled.  We do not have any other solution right now that would provide the same level of user experience.

kdudka (approved):

siteshwar:
I am merging this for now. But we may have to remove this check in future.

PR #27, opened 2023-06-27
ci: Use older version of Ubuntu to run podman compose

siteshwar:
An update in the podman package in latest Ubuntu has caused broken network functionality, which is causing failures in the CI.

These changes are a workaround for this bug and should be reverted when this bug has been fixed.

Related: https://bugs.launchpad.net/ubuntu/+source/libpod/+bug/2024394


kdudka (approved):
The CI appears fixed!

PR #29, opened 2023-06-28
docs: add `MacOS Users` section in the development documentation

jperezdealgaba:
I have added the setup for MacOS (M1 processors).

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -16,14 +16,14 @@ By default, this script will deploy a `postgresql` instance, the worker and the 
 
 You can skip deployment and basic settings if you use the script.
 
-### MacOS Users
+### macOS Users
 
 If you are using MacOS to develop for OpenScanHub you will need to set up a different environment:
 
 1. You will need to install the latest Bash version using brew:
     `brew install bash`
 2. Locate the new package:
-   `which bash` -> `/opt/homebrew/bin/bash`
+   `command bash` -> `/opt/homebrew/bin/bash`
 3. Run it:
     `/opt/homebrew/bin/bash`
 4. In the new console, you can run the scripts form the [Using scripts](#using-scripts) section.
--- a/docs/development.md
+++ b/docs/development.md
@@ -16,6 +16,34 @@ By default, this script will deploy a `postgresql` instance, the worker and the 
 
 You can skip deployment and basic settings if you use the script.
 
+### MacOS Users
+
+If you are using MacOS to develop for OpenScanHub you will need to set up a different environment:
+
+1. You will need to install the latest Bash version using brew:
+    `brew install bash`
+2. Locate the new package:
+   `which bash` -> `/opt/homebrew/bin/bash`
+3. Run it:
+    `/opt/homebrew/bin/bash`
+4. In the new console, you can run the scripts form the [Using scripts](#using-scripts) section.
+
+Once it has been set up, you will need update the mock config:
+
+1. Create an [admin user](#osh-hub-users)
+2. Login as the admin user at: http://localhost:8000/auth/login/?next=/
+3. On the top right corner, click on the `admin` section
+4. Go to `Scan` section -> `Mock configs` -> `Add`:
+    - Name: `fedora-37-aarch64`
+    - Enabled: `true`
+5. Click on `Save`
+6. Go back to `Mock configs`, look for the recent created one and take its ID.
+7. Go back to the `admin` section -> `Scan` section -> `Tags` -> `Fedora 37`.
+8. Click on `Edit` and change the `Mock Config ID` for the one of step 6.
+9. Click on `Save`
+
+You are ready to start scanning!
+
 ## OSH worker
 
 Worker depends on some system packages not available from PyPI, needs to run under root user and has kinda complex setup which are all the reasons to run it in a container.


docs/development.md
```diff
@@ -191,19 +184,19 @@ osh/hub/scripts/osh-xmlrpc-client.py \
...
 
- 1. Code execution starts in client, for a specific command, e.g. [diff-build](https://gitlab.cee.redhat.com/covscan/covscan/blob/master/osh/client/commands/cmd_diff_build.py).
-   * Files are uploaded to server via [`upload_file` XML-RPC call](https://gitlab.cee.redhat.com/covscan/covscan/blob/master/osh/client/commands/shortcuts.py#L104).
-   * The XML-RPC call itself [is defined](https://gitlab.cee.redhat.com/covscan/covscan/blob/master/osh/hub/settings.py#L160) [in kobo](https://github.com/release-engineering/kobo/blob/master/kobo/django/upload/xmlrpc.py#L19).
- 2. Server code path starts in XML-RPC API at specific method for particular scan type, e.g. for [mock builds](https://gitlab.cee.redhat.com/covscan/covscan/blob/master/osh/hub/osh_xmlrpc/scan.py#L57).
- 3. There is a hierarchical structure for configuring data for scan in `osh/hub/errata/scanner.py`, for client scans this is [ClientScanScheduler](https://gitlab.cee.redhat.com/covscan/covscan/blob/master/osh/hub/errata/scanner.py#L374).
+ 1. Code execution starts in client, for a specific command, e.g. [diff-build](https://github.com/openscanhub/openscanhub/blob/main/osh/client/commands/cmd_diff_build.py).
```
siteshwar: Please amend your previous commit and remove the changes of adding internal GitLab links from it.

jperezdealgaba: Check the new commit please @siteshwar . I already fixed that

siteshwar: I still see https://github.com/openscanhub/openscanhub/pull/29/commits/2e8d41a343276c0afb5f028c61c3fa3db21d2778 in your PR which adds links to internal GitLab.

jperezdealgaba: The last commit is [6857cf4](https://github.com/openscanhub/openscanhub/pull/29/commits/6857cf4136909c2719753818e78da4b18b62f3e9)

siteshwar: You are going back and forth on changes. First you added references to internal links through https://github.com/openscanhub/openscanhub/pull/29/commits/2e8d41a343276c0afb5f028c61c3fa3db21d2778 and later switched them back to external links by https://github.com/openscanhub/openscanhub/pull/29/commits/6857cf4136909c2719753818e78da4b18b62f3e9. The changes in the first commit are unnecessary and should be amended.

jperezdealgaba: Sorry, but I am not understanding it. The last commit (https://github.com/openscanhub/openscanhub/pull/29/files) is the original file with just my new section on it.

jperezdealgaba: All referenced links there are from github

kdudka: @jperezdealgaba As I understand it, @siteshwar wanted you to squash the commits.  I have done it on your behalf.

kdudka:
@jperezdealgaba I have pushed a few fixup commits to fix the white-spaces and improve rendering on GitHub:
- original: https://github.com/openscanhub/openscanhub/blob/7a584f53684196ebac49f40ced5e54f06fe8ea02/docs/development.md#macos-users
- updated: https://github.com/openscanhub/openscanhub/blob/960b89b11d24d37fb412c6181daeb7cd2c84994e/docs/development.md#macos-users

jperezdealgaba:
Thanks for that Kamil!

kdudka (approved):
Looks good to me.  Thanks for the contribution!

docs/development.md
```diff
@@ -16,6 +16,34 @@ By default, this script will deploy a `postgresql` instance, the worker and the
...
+2. Login as the admin user at: http://localhost:8000/auth/login/?next=/
+3. On the top right corner, click on the `admin` section
+4. Go to `Scan` section -> `Mock configs` -> `Add`:
+    - Name: `fedora-37-aarch64`
+    - Enabled: `true`
+5. Click on `Save`
+6. Go back to `Mock configs`, look for the recent created one and take its ID.
+7. Go back to the `admin` section -> `Scan` section -> `Tags` -> `Fedora 37`.
+8. Click on `Edit` and change the `Mock Config ID` for the one of step 6.
+9. Click on `Save`
```
lzaoral, line 43: OSH should support aarch64 buildroots out of the box but it looks like we missed Fedora 37.  This is actually a bug that should be fixed in https://github.com/openscanhub/openscanhub/blob/main/osh/hub/scan/fixtures/initial_data.json instead.

docs/development.md
```diff
@@ -16,6 +16,34 @@ By default, this script will deploy a `postgresql` instance, the worker and the
...
 You can skip deployment and basic settings if you use the script.
 
+### MacOS Users
+
+If you are using MacOS to develop for OpenScanHub you will need to set up a different environment:
+
+1. You will need to install the latest Bash version using brew:
+    `brew install bash`
+2. Locate the new package:
+   `which bash` -> `/opt/homebrew/bin/bash`
```
lzaoral, line 26:
```suggestion
   `command -v bash` -> `/opt/homebrew/bin/bash`
```
I'd use a portable solution that does not require any extra programs like `which` instead.  I don't know if `which` installed by default on macOS but it definitely is not the case with some Linux distros.  Also, `command` is specified by POSIX: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/command.html

docs/development.md
```diff
@@ -16,6 +16,34 @@ By default, this script will deploy a `postgresql` instance, the worker and the
...
+3. Run it:
+    `/opt/homebrew/bin/bash`
+4. In the new console, you can run the scripts form the [Using scripts](#using-scripts) section.
```
lzaoral, line 29: Hmm, AFAIK, all scripts have a shebang to execute `bash` through `env` so this step should not be necessary.  Could please check?

docs/development.md
```diff
@@ -16,6 +16,34 @@ By default, this script will deploy a `postgresql` instance, the worker and the
 
 You can skip deployment and basic settings if you use the script.
 
+### MacOS Users
+
+If you are using MacOS to develop for OpenScanHub you will need to set up a different environment:
```
lzaoral, line 21: Nitpick: `s/MacOS/macOS` as that seem to be name capitalisation that Apple uses: https://www.apple.com/macos/ventura/

lzaoral (changes requested):
Thank you @jperezdealgaba!  Some suggestions inline.

docs/development.md
jperezdealgaba: I had to do this because if you just execute the script directly, the terminal you are using will be used instead of the recently installed. In this case, zsh and the default macOS terminal (GNU Bash version 3.2) doesn't support the "mapfile" function and doesn't let the script run. That's why I added these extra steps

docs/development.md
jperezdealgaba: I updated this in new commit

lzaoral: Resolved in 82f4f94.

docs/development.md
lzaoral: This is still not resolved because `command` is not the same as `command -v`. See the linked POSIX manual page in comment https://github.com/openscanhub/openscanhub/pull/29#discussion_r1255760844.

docs/development.md
lzaoral: Just to prevent confusion, by terminal you mean a shell, right?  There shouldn't be any difference between the output of `which` and the binary that will be used because our shebangs invoke `bash` through `env` (e.g. #!/usr/bin/env bash) so the step 3 still seems redundant to me.  

Nevertheless, I believe that there is no real reason to use `mapfile` and, thus, force macOS users to install a newer version of `bash` from Homebrew.  I'll try to simplify the script a bit a I'll let you know when it's ready to be tested.  Thank you!

PR #34, opened 2023-06-29
Enable Dependabot

jamacku:
Dependabot is a GitHub dependency manager that automatically updates project dependencies. This commit adds configuration for updating GitHub Actions.

follow-up to discussion in #25


kdudka (approved):

siteshwar:
LGTM. In future, we may look into adding other package ecosystems than GitHub Actions.

PR #35, opened 2023-06-29
hub: waiving/reporting: bugzilla: use base link on settings

isinyaaa:
Closes: https://gitlab.cee.redhat.com/covscan/covscan/-/issues/276

### DIFF
--- a/osh/hub/settings_local.py
+++ b/osh/hub/settings_local.py
@@ -101,7 +101,7 @@ TASK_DIR = os.path.join(FILES_PATH, 'tasks')
 # Root directory for uploaded files
 UPLOAD_DIR = os.path.join(FILES_PATH, 'upload')
 
-BZ_URL = "https://bugzilla.stage.redhat.com/xmlrpc.cgi"
+BZ_URL = "https://bugzilla.stage.redhat.com"
 BZ_API_KEY = "xxxxxx"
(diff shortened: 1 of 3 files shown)


lzaoral:
Reassigning reviewers because @kdudka is busier than usual.

lzaoral (approved):
LGTM!  Thank you, @isinyaaa!

osh/hub/settings_local.py
```diff
@@ -101,7 +101,7 @@
 # Root directory for uploaded files
 UPLOAD_DIR = os.path.join(FILES_PATH, 'upload')
 
-BZ_URL = "https://bugzilla.stage.redhat.com/xmlrpc.cgi"
+BZ_URL = "https://bugzilla.stage.redhat.com"
```
rhyw, line 104: do you think it's necessary to also update stage/prod settings in `osh-rh-conf` as well for consistency?

since `get_client` uses urljoin, leaving it unchanged won't do any harm.

rhyw: Other than that, everything looks good.

lzaoral: Yes, the production configuration must be updated as well.

lzaoral: I've prepared a MR with related modifications to `osh-rh-conf`.

PR #36, opened 2023-06-30
packaging: create the /var/log/osh directory

kdudka:
... when installing the worker or hub packages.  We use systemd to write the worker log now and systemd does not automatically create the directory (unlike the osh-worker daemon).

Fixes: commit 30fadad7945a4c1de316956a77641b93a99bd649

### DIFF
--- a/osh.spec
+++ b/osh.spec
@@ -193,6 +193,7 @@ rm -rf %{buildroot}%{python3_sitelib}/scripts
 %{python3_sitelib}/osh/worker
 %{_unitdir}/osh-worker.service
 %attr(754,root,root) /usr/sbin/osh-worker
+%dir %attr(775,root,root) /var/log/osh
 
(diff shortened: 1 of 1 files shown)


osh.spec
```diff
@@ -219,6 +220,7 @@ fi
 %{_unitdir}/osh-stats.*
 %exclude %{python3_sitelib}/osh/hub/settings_local.py*
 %exclude %{python3_sitelib}/osh/hub/__pycache__/settings_local.*
+%dir %attr(775,root,root) /var/log/osh
```
siteshwar, line 223: Why do we need to create it on the hub?

kdudka: Otherwise the `/var/log/osh` would be unowned if we installed `osh-hub` only.

siteshwar (approved):

PR #48, opened 2023-07-11
containers: Simplify and `deploy.sh`

lzaoral:
* Simplifies the code.
* Removes constructs not compatible with default Bash on macOS (e.g. `mapfile`)
* Fixes the `--clean` option on Linux.
* Fixes the `--force` option on Linux.


lzaoral:
@jperezdealgaba  Can you please test that these changes make the `deploy.sh` script work on macOS without any extra dependencies and if the `--clean` option works as well?  Thanks in advance!

jperezdealgaba:
Hey! 
The deploy.sh seems that it is working fine and also the --clean flag!. Just to make sure: When I started the environment using the scripts, I used the init-db.sh script, not the deploy.sh script. Should I now use that one?

kdudka (approved):
Perfect.  Thank you both!

I spotted a typo in a commit message: `fix compose cleanup then --force is used` -> `fix compose cleanup when --force is used`

lzaoral:
> I spotted a typo in a commit message: `fix compose cleanup then --force is used` -> `fix compose cleanup when --force is used`

Addressed in 31337d2eb70c7a4845a45d7a1630114257037fdd.

PR #49, opened 2023-07-11
hub: add a fixture for the `fedora-37-aarch64` mock config

lzaoral:
So that users can use the default configuration also on aarch64 machines.

Related: https://github.com/openscanhub/openscanhub/pull/29/files#r1255756164


kdudka (approved):
Looks good.  Thanks!

Issue #32, opened 2023-06-29
ci: enable Coverity integration in the upstream CI


kdudka:
It would be nice to use Coverity (the scanning service freely available for open-source project) in the upstream CI of OpenScanHub.

lzaoral:
@kdudka Would the integration like we have in @csutils be sufficient?  At the moment, there is not an easy way to do such scans from forks because the scans are bound to a single secret token.  Moreover, even if it was possible, it would drain out our daily scan quota quite fast.

lzaoral:
In the meantime, I've enabled the integration on Coverity's side.

kdudka:
Yes, scanning the `main` branch should be sufficient for now.  Later on we can automate scanning of OpenScanHub source code with OpenScanHub so that we do not need to integrate each scanner separately.

lzaoral:
Reopening.  While the GitHub Actions workflow is set-up correctly but the build itself still needs some tweaking:
```
Last Build Status: Failed. Your build has failed due to the following reason. Please fix the error and upload the build again.
Error details: The build uploaded has been only partially compiled. We recommend at least 85% capture success to avoid false-positives during analysis. As per last few lines of cov-int/build-log.txt, the percentage of compilation units ready for analysis is 75% which is less than the expected 85%...
```

lzaoral:
The job log:
```
Coverity Build Capture (64-bit) version 2022.12.2 on Linux 5.15.0-1041-azure x86_64
Internal version numbers: 126f4dac91 p-2022.12-push-76


[STATUS] Running filesystem capture search...
[STATUS] Emitting 157 files from filesystem capture
|0----------25-----------50----------75---------100|
****************************************************
Warning:  Emitted 3 JavaScript compilation units (75%) successfully
Emitted 124 Python compilation units (100%) successfully

3 JavaScript compilation units (75%) are ready for analysis
124 Python compilation units (100%) are ready for analysis
The cov-build utility completed successfully.
```

lzaoral:
Both mine and @kdudka's machines produce a different output when `cov-build` is run locally:
```console
$ cov-build --dir cov-int --no-command --fs-capture-search ./osh
Coverity Build Capture (64-bit) version 2022.12.2 on Linux 6.4.2-arch1-1 x86_64
Internal version numbers: 126f4dac91 p-2022.12-push-76


[STATUS] Running filesystem capture search...
[STATUS] Emitting 268 files from filesystem capture
[...]

PR #51, opened 2023-07-12
ci: enable scanning with Coverity

lzaoral:
At the moment, it is only possible to scan the `main` branch.  This PR is exceptionally done from the repository itself to test that all other Coverity related settings are configured correctly.

Resolves: https://github.com/openscanhub/openscanhub/issues/32

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -1,0 +1,25 @@
+---
+name: Coverity
+
+on:
+  push:
+    branches: [main]
+
+jobs:
+  coverity:
+    if: github.repository == 'openscanhub/openscanhub'
+    name: Coverity Scan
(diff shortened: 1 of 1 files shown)


lzaoral:
Actually, I'm blind.  The workflow has to be push `main` only so it does not really matter where is the source branch located.  Sorry for the confusion!

siteshwar:
We should continue to use full URLs while referencing issues to avoid confusion with internal GitLab issues.

lzaoral:
@siteshwar That's what I did but GitHub just mangled that to its own format anyway (try to edi the comment to see the actual URL is there).  While I can enclose it in e.g. backticks, I would also like to retain the automatic linking with the issue from this PR.

kdudka (approved):
Looks good.  Thanks!

PR #56, opened 2023-07-12
hub: remove dead code from `waiving/service.py`

lzaoral:
Detected by Coverity.


lzaoral:
Actually, it looks like some stuff from `osh/hub/templates/waiving/result.html` can be removed as well.  I'll update the PR shortly.

lzaoral:
Blocked by #58.

PR #58, opened 2023-07-13
ci: Disable patch check in codecov

siteshwar:
CI should not fail if a line changed through a patch is not covered by tests.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,3 +1,6 @@
 # When modifying this file, please validate using
 # curl -X POST --data-binary @codecov.yml https://codecov.io/validate
 comment: false
+coverage:
+  status:
+    patch: off


jamacku (commented):
I don't think this is the best solution, you will lose valuable information about the impact of your change on test coverage.

I think the better approach would be to set [threshold](https://docs.codecov.com/docs/commit-status#threshold) instead.

kdudka:
The problem is that we are not able to capture coverage for some areas of the code because python-coverage does not work well with sub-processes: https://coverage.readthedocs.io/en/coverage-5.1/subprocess.html#measuring-sub-processes  This may lead to misleading results and artificial CI failures.

siteshwar:
> I think the better approach would be to set [threshold](https://docs.codecov.com/docs/commit-status#threshold) instead.

It can confuse new contributors due to inconsistent CI failures (as mentioned by @kdudka), but it could be a good idea for the future.

jamacku (approved):
LGTM :+1:

kdudka (approved):

PR #61, opened 2023-07-18
Restrict user list/detail view to authenticated user only

rhyw:
The users page is rendered in kobo, thus changing is a little bit tricky. The patch attempts to hijack the original view through a middleware such that we don't have to change code in kobo.

Fixes OSH-78


osh/hub/other/middleware.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def __call__(self, request):
+        path_info = request.META["PATH_INFO"]
+        # This `user/list` comes from kobo's view name for user
+        kobo_user_view_url = reverse("user/list")
+        if path_info.startswith(kobo_user_view_url):
+            if not request.user.is_authenticated:
+                template = get_template("base.html")
+                context = {
+                    "info_message": "Please login to view users.",
```
rhyw, line 27: Note: This allows to show a banner message in case of anonymous access. It's from `templates/layout.html` in kobo.

lzaoral:
Thank you for your contribution, @rhyw!

As you said youself, updating `kobo` would be more straightforward and less hacky.  Could you please ask upstream first, if they would be fine with such change?  Either directly or by making this feature, e.g. configurable in `settings.py`.  Thank you!

rhyw:
@lzaoral thanks for the review, good point. I filed https://github.com/release-engineering/kobo/issues/209 .

rhyw:
The issue filed in kobo seem to be unresponsive lately.

I've updated the patch to allow access only to admins, along with test cases that covers the scenarios.

After a quick look at kobo source, to fix the issue upstream, I think it requires some dedicated changes including for example to either
- implement a **conditional** permission control on the shared `ExtraListView` that only applies to "users/list".
- Or to add a seperate view specifically for users, instead of relying on the shared view.

Either way seem to require some tweaks in the `MenuItem`. The changes may break kobo's existing test cases I assume, not sure about the effort though.

That said, I prefer to merge this MR for now. In case there's updates from kobo, we could act upon. I personally could submit a patch in future, but it's not a priority for me right now, also I'd like to get some insights from kobo maintainers before proceeding.

Thoughts? @kdudka @lzaoral

lzaoral:
@rhyw Sorry for a late reply.  I'd wait a bit longer for the upstream response.  `kobo` is primarily maintained by a single person and none of our PRs or issues from the past few weeks got a response yet.

kdudka:
We are currently facing some issues with the changes not taking an effect on the internal staging OSH instance.  I am marking this pull request as a draft until the root cause of the issues is identified.

lzaoral:
@kdudka I still think that making this change directly in `kobo` would be cleaner solution.

rhyw:
Closing in favor of https://github.com/openscanhub/openscanhub/pull/146

Thanks for all the reviews. I failed to find out why it won't work after deployed. I've proposed a separate patch upstream, hopefully to resolve the issue without having to hijacking the views through middlewares.

PR #62, opened 2023-07-18
hub: improve secret handling

isinyaaa:
Closes: https://gitlab.cee.redhat.com/covscan/covscan/-/issues/241

### DIFF
--- a/osh/hub/settings.py
+++ b/osh/hub/settings.py
@@ -19,6 +19,9 @@ URL_PREFIX = "/osh"
 # file to read the real SECRET_KEY from
 SECRET_KEY_FILE = "/var/lib/osh/hub/secret_key"
 
+# where to read API keys from
+SECRETS_DIR = "/etc/osh/hub/secrets"
+
 # Language code for this installation. All choices can be found here:
 # http://www.i18nguy.com/unicode/language-identifiers.html
 LANGUAGE_CODE = 'en-us'
@@ -185,6 +188,18 @@ try:
     from .settings_local import *  # noqa
 except ImportError:
     pass
+
+
+def _get_secret(name):
+    try:
+        with open(os.path.join(SECRETS_DIR, name)) as f:
+            return f.read().strip()
+    except OSError:
+        return None
+
+
+BZ_API_KEY = _get_secret('bugzilla_secret')
+JIRA_API_KEY = _get_secret('jira_secret')
 
 # read the real SECRET_KEY from SECRET_KEY_FILE if availble
 try:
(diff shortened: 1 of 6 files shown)


lzaoral (approved):
LGTM! thanky ou, @isinyaaa!  Let's wait for @kdudka's review, though.

osh/hub/settings.py
```diff
@@ -19,6 +19,9 @@
 # file to read the real SECRET_KEY from
 SECRET_KEY_FILE = "/var/lib/osh/hub/secret_key"
 
+# where to read API keys from
+SETTINGS_DIR = "/etc/osh/hub/secrets"
```
kdudka, line 23: I would name it `SECRETS_DIR` or `SECRET_SETTINGS_DIR` to distinguish it from other directories that contain configuration files.

osh/hub/settings.py
```diff
@@ -186,6 +189,18 @@
 except ImportError:
     pass
 
+
+def _get_settings(name):
```
kdudka, line 193: Same here.   I would name it `_get_secret()`.

osh/hub/settings_local.py
```diff
@@ -11,6 +11,8 @@
 
 DEBUG = True
 
+SETTINGS_DIR = ".secrets"
```
kdudka, line 14: There should be the same comment above the line as we have in `settings.py`.

kdudka (changes requested):
Sorry for the delay!  Overall looks good.  Some minor remarks inline.

osh/hub/settings.py
```diff
@@ -186,6 +189,18 @@
 except ImportError:
     pass
 
+
+def _get_secret(name):
+    try:
+        with open(os.path.join(SECRETS_DIR, name)) as f:
+            return f.read().strip()
+    except OSError:
```
lzaoral, line 197: @kdudka Should we also silently skip permission related errors or just missing files?

kdudka: Assigning `None` (at least in case of missing files) sounds fine to me.  The hub will continue to work until the secret is actually needed.  Not everybody uses Bugzilla or Jira.  But we should probably return `None` explicitly in this case.

lzaoral: But `None` is returned even now, though not explicitly.  (Any Python function/method without explicit `return` statement returns `None`.)

isinyaaa: oh you're right, either way it's clearer to write it out I guess.

PR #63, opened 2023-07-18
ci: remove `.gitlab-ci.yml`

lzaoral:
This file is practically obsolete and, moreover, contains some URLs not accessible from the outside.


kdudka (approved):
Good catch!

PR #69, opened 2023-07-20
hub: refactor resubmit scan

lzaoral:
See commit messages for details.

This is only a **draft** because these changes are untested.


lzaoral:
Note that this PR mainly focuses on readability!  The rescan feature still does not work properly.

PR #70, opened 2023-07-21
ci: remove scripts used by the old GitLab CI

lzaoral:
They are no longer used by any other scripts or mentioned on documentation.


lzaoral:
This PR is wrong.  Apparently, I haven't executed `git grep` in the repository root...

PR #79, opened 2023-07-28
client: do not crash on unknown profile

rhyw:
Fixes OSH-145

### DIFF
--- a/osh/client/commands/cmd_version_diff_build.py
+++ b/osh/client/commands/cmd_version_diff_build.py
@@ -15,7 +15,8 @@ from .common import (add_analyzers_option, add_comment_option,
                      add_priority_option, add_profile_option,
                      add_task_id_file_option)
 from .shortcuts import (check_analyzers, handle_perm_denied, upload_file,
-                        verify_koji_build, verify_mock)
+                        verify_koji_build, verify_mock,
+                        verify_scan_profile_exists)
 
 
 class Version_Diff_Build(osh.client.OshCommand):
@@ -198,6 +199,9 @@ is not even one in your user configuration file \
                 self.parser.error(str(ex))
             options['analyzers'] = analyzers
         if profile:
+            result = verify_scan_profile_exists(self.hub, profile)
+            if result is not None:
+                self.parser.error(result)
             options['profile'] = profile
 
         if brew_build:
--- a/osh/client/commands/shortcuts.py
+++ b/osh/client/commands/shortcuts.py
@@ -145,3 +145,11 @@ def upload_file(hub, srpm, target_dir, parser):
         return hub.upload_file(os.path.expanduser(srpm), target_dir)
     except Fault as e:
         handle_perm_denied(e, parser)
+
+
+def verify_scan_profile_exists(hub, profile_name):
+    """Verify if scan profile exists"""
+    profiles = hub.scan.list_profiles()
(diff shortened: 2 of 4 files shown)


osh/client/commands/cmd_version_diff_build.py
```diff
@@ -198,7 +199,10 @@ def run(self, *args, **kwargs):  # noqa: C901
                 self.parser.error(str(ex))
             options['analyzers'] = analyzers
         if profile:
-            options['profile'] = profile
+            result = verify_scan_profile_exists(self.hub, profile)
+            if result is not None:
+                self.parser.error(result)
+        options['profile'] = profile
```
lzaoral, line 205: Same as above.

osh/client/commands/shortcuts.py
```diff
@@ -145,3 +145,12 @@ def upload_file(hub, srpm, target_dir, parser):
         return hub.upload_file(os.path.expanduser(srpm), target_dir)
     except Fault as e:
         handle_perm_denied(e, parser)
+
+
+def verify_scan_profile_exists(hub, profile_name):
+    """Verify if scan profile exists"""
+    profiles = hub.scan.list_profiles()
+    profile_exists = any(p["name"] == profile_name for p in profiles)
+    if profile_exists:
```
lzaoral, line 154: There is no need to introduce a local variable that will be set and read only once.
```suggestion
    if any(p["name"] == profile_name for p in profiles):
```

lzaoral (changes requested):
Thank you for your contribution!  See comments below.

Please, amend the commit messages to contain the following string:
```
Resolves: <fill link to OSH-145 Jira ticket>
```

rhyw:
> Thank you for your contribution! See comments below.
> 
> Please, amend the commit messages to contain the following string:
> 
> ```
> Resolves: <fill link to OSH-145 Jira ticket>
> ```

I've been working on other RH internal projects as well. TBH I'm not quite comfortable pasting internal links which were not accessible from outside of the world.

osh/client/commands/shortcuts.py
rhyw: I originally also want a one liner. but some may say one liner ruins readability.

Anyway, I'm ok with getting rid of this accidentally introduced local vars. Thanks for the suggestion.

osh/client/commands/cmd_version_diff_build.py
rhyw: Fixed.

rhyw:
> Resolves: <fill link to OSH-145 Jira ticket>

We do have conventions in historical commits. I've amended the commit message to follow that! Thanks for the suggestion.

lzaoral (changes requested):
Perfect, thank you!  I see that the coverage decreased a bit.  Could you please a modify the first `mock-build` task in `containers/scripts/generate_integration_test_coverage.sh` to include the `--profile=default` option?  Looking at the codecov logs, we not have this option covered yet.  The behaviour should be the same because `default` is the implicit profile used by `osh-cli`.

rhyw:
Updated as per suggestion.

Code coupled too much with kobo is the biggest challenge to me. have to traverse repo back and forth to get an idea how different modules work together.

lzaoral (approved):
LGTM!  Thank you, @rhyw!

PR #80, opened 2023-07-28
containers: fix incorrect mail setup while creating user in init-db.sh

rhyw:


rhyw:
Now working on getting rid of the user setup by openning a shell and run commands inside.

lzaoral:
Thank you, @rhyw!  I believe we can remove that for loop completely.  @kdudka Can you confirm that you did not use this account during development in containers?

rhyw:
I think we still need some pre populated users for dev purposes. I removed the shell stuff and use a fixture instead:
https://github.com/openscanhub/openscanhub/pull/81

kdudka (approved):
Thanks!  That was obviously a typo.  The prefix should be `containers: ` (with the trailing `s`) according to https://github.com/openscanhub/openscanhub/blob/main/CONTRIBUTING.md

lzaoral:
> I think we still need some pre populated users for dev purposes. I removed the shell stuff and use a fixture instead: https://github.com/openscanhub/openscanhub/pull/81

I've approved this PR because it fixes an obvious bug and I'll move the discussion about the user list for development to #81.

lzaoral (approved):

rhyw:
commit message prefix amended.

PR #94, opened 2023-08-16
hub: Make footer configurable in base template

siteshwar:
Resolves: https://gitlab.cee.redhat.com/covscan/covscan/-/issues/174

### DIFF
--- a/osh/hub/settings.py
+++ b/osh/hub/settings.py
@@ -180,6 +180,8 @@ ET_SCAN_PRIORITY = 20
 
 VALID_TASK_LOG_EXTENSIONS = ['.log', '.ini', '.err', '.out', '.js', '.txt']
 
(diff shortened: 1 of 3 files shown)


siteshwar:
@lzaoral The changes in this pull request work fine for index page, but we need to find a way to pass `settings.BASE_HTML_FOOTER` variable to rest of the pages that are rendered through django.

lzaoral:
What about using a separate template for the footer?  In that case, anyone can override it how they please.

siteshwar:
>  What about using a separate template for the footer? In that case, anyone can override it how they please.

@kdudka Do you agree with this?

siteshwar:
> Yes, that was exactly my idea. We can include the footer in the same way as we include index-content.html in osh/hub/templates/index.html and put overrides to osh-rh-conf.


Ok, I will open a separate pull request to follow up on it.

Issue #111, opened 2023-08-28
hub: add some checker groups to initial fixtures 


lzaoral:
We should ship initial fixtures with some basic checker groups added, if possible.  Otherwise, all results will be marked as `Unsorted`: ![](https://user-images.githubusercontent.com/48823770/263649388-725930b8-51f5-428a-b4e9-39a43fb92f7b.png)

kdudka:
The legacy checker groups we use internally are specific to Coverity.  The most interesting checker groups are the ones that were added recently for the open-source scanners and those should be upstreamed:

- `Compiler Warnings` for `COMPILER_WARNING`
- `Cppcheck Warnings` for `CPPCHECK_WARNING`
- `Shell` for `SHELLCHECK_WARNING`
- `GCC Analyzer` for `GCC_ANALYZER_WARNING`
- `Clang` for `CLANG_WARNING`
- `Hard-coded Secrets` for `GITLEAKS_WARNING`

PR #112, opened 2023-08-28
hub: update fixtures

lzaoral:
* enable smatch in fixtures
* update fixtures to only include publicly available products and mock configs

### DIFF
--- a/containers/hub/run.sh
+++ b/containers/hub/run.sh
@@ -20,7 +20,7 @@ osh/hub/manage.py migrate
 # hub work.
 if [ "$(osh/hub/manage.py dumpdata scan.MockConfig)" = "[]" ]; then
     osh/hub/manage.py loaddata \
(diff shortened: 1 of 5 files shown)


kdudka (approved):
This works for me but let's double check whether all the changes are compatible with the public OSH deployment that @siteshwar is working on.

lzaoral:
@kdudka Definitely!  In the meantime, I might as well add the fixtures for the suggested checker groups in #111.

lzaoral:
Hmm, the `multi` profile fixture should be updates as well to include `valgrind` and `smatch` and `gcc-unstable` should be removed as that seems to be just a remnant of some experiment.

lzaoral:
... and there is no profile for `pylint` now.

siteshwar (approved):

kdudka (approved):

PR #115, opened 2023-08-29
osh: remove internal gitlab links

lzaoral:


scripts/kill_django_server.sh
```diff
@@ -10,8 +10,3 @@ django_server_coverage_pid=$(ps -eo pid,cmd | grep -i runserver | grep -v 'grep'
 kill -15 "$django_server_coverage_pid"
 # Wait will print an error if process has been already killed
 wait "$django_server_coverage_pid" 2>/dev/null || :
-
-# Clean up tasks directory in the container. This has to be done here as it may cause issues with
-# starting new jobs in GitLab CI if the user that runs the job in container is different from host
-# user
-rm -rvf osh/hub/tasks
```
siteshwar, line 17: It should be safe to keep these lines even if we do not use GitLab CI and the mentioned condition is true in any other environment.

lzaoral: Ok, I'll just remove the GitLab reference then.

lzaoral: Addressed in d52c640.

siteshwar (approved):

PR #119, opened 2023-08-30
hub: remove `errata/service.py`

lzaoral:
As discussed with @kdudka in person, the rescan functionality in the admin interface was never actively used and is currently broken.  Therefore, it should be removed with the whole file because the rest of the code is dead.


kdudka (approved):
Looks good.  Thanks!

PR #122, opened 2023-09-01
hub: fix overlapping pagination controls

lzaoral:
Kobo's CSS declare the `.paginator` class as `float`.  Therefore, in some cases the pagination controls would overlap with other page content making the visuals look corrupted.

Override the broken style in OSH because changes in kobo's styles may break other projects that depend on them.

Resolves: https://gitlab.cee.redhat.com/osh/covscan/-/issues/227

Testing instructions:
* Create a local OSH instance with the database snapshot restored.
* Check that all views with pagination controls are no longer corrupted with this PR appleid.

### DIFF
--- a/osh/hub/static-assets/css/redhat.css
+++ b/osh/hub/static-assets/css/redhat.css
@@ -782,3 +782,14 @@ div#defects div.events_list div.code {
 }
 
 /**/
+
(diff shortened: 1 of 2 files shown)


jperezdealgaba (approved):
Tried it and /lgtm

PR #123, opened 2023-09-04
hub: add missing view titles

lzaoral:


lzaoral:
Depends on https://github.com/release-engineering/kobo/pull/227.

kdudka (approved):
Looks good.  Just to be sure: The views will not be more broken than before if we use the updated OSH code in combination with an old version of kobo, right?

lzaoral:
The exact opposite, they will still look better!  All commit messages that mention some dependency on `kobo` just do nothing when used with old `kobo`.

PR #135, opened 2023-09-06
create-release.sh: create a release on GitHub

kdudka:
The script is taken from csutils/csdiff:
https://github.com/csutils/csdiff/blob/csdiff-3.0.4/upload-release.sh

But the script does not upload any files.  It only creates the release on GitHub so that we can use more native download URLs.

### DIFF
--- a/create-release.sh
+++ b/create-release.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+set -e
+set -x
+SELF="$0"
+TAG="$1"
+TOKEN="$2"
+
+PROJ="openscanhub/openscanhub"
+
+usage() {
+    printf "Usage: %s TAG TOKEN\n" "$SELF" >&2
+    exit 1
+}
+
+# check arguments
+test "$TAG" = "$(git describe --tags "$TAG")" || usage
+test -n "$TOKEN" || usage
+
+# create a new release on GitHub
+curl "https://api.github.com/repos/${PROJ}/releases" \
+    --fail --verbose \
+    --header "Authorization: token ${TOKEN}" \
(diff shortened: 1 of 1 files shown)


kdudka:
@siteshwar I have created the script as we discussed on the meeting. Unfortunately, it does not seem to do anything useful: https://github.com/openscanhub/openscanhub/releases/tag/osh-0.9.4

The download URLs are the same as if it was a git tag only.  I think the only way to get the URLs you want is to **upload** the release tarballs to GitHub, like we do for `csutils` projects.  In this case, I will need to upload the signatures, too, because then users need to trust me (and my build environment) as creator of the tarballs, rather than relying on GitHub to create the tarballs out of a signed git tag.

frenzymadness:
I'm not sure I'm the right person to review this as I don't understand the intention here. The vast majority of projects I work with or maintain use tags. Usually, I just tag a commit and then push it. Then we use the URL pointing to the newly created tag in specfiles.

kdudka:
@frenzymadness Thanks for feedback!  Just to be clear, could you please confirm that the following URL is acceptable for the `Source:` tag in the spec file of a Fedora package?
https://github.com/openscanhub/openscanhub/archive/refs/tags/osh-0.9.4.tar.gz

frenzymadness:
It looks like more than 150 packages use the same schema so I think it's acceptable.

kdudka:
@siteshwar Are you fine with closing this without merging then?

siteshwar:
> @siteshwar Are you fine with closing this without merging then?

Yes, we can close this.

PR #139, opened 2023-09-13
Small refactoring

lzaoral:
This PR contains vaerious small changes that were supposed to be part of something bigger.  I've formed them into something mergeable instead because I'm no longer going to work on OSH starting next month.


kdudka (approved):
Looks good.  Thanks!

PR #140, opened 2023-09-13
packaging: Package the example httpd config

lzaoral:


osh.spec
```diff
@@ -254,6 +257,7 @@ pg_isready -h localhost && %{python3_sitelib}/osh/hub/manage.py migrate
 %files hub-conf-devel
 %attr(640,root,apache) %config(noreplace) %{python3_sitelib}/osh/hub/settings_local.py
 %attr(640,root,apache) %config(noreplace) %{python3_sitelib}/osh/hub/__pycache__/settings_local*.pyc
+%attr(640,root,apache) %config(noreplace) %{_sysconfdir}/httpd/conf.d/osh-hub-httpd.conf
```
kdudka, line 260: IIRC, according to Fedora Packaging Guidelines the package should either own the directory where it installs the file, or depend on a package that installs the directory.  `osh-hub` depends on `httpd` but `hub-conf-devel` can be installed without `osh-hub`.

kdudka (approved):

osh.spec
lzaoral: Hmm, in that case it should own `%{python3_sitelib}/osh` as well, right?

lzaoral: Addressed in 275f8f1 and a9b2bda.

osh.spec
```diff
@@ -107,17 +107,23 @@ Obsoletes: covscan-hub < %{version}
...
-%description ${sub}-conf-devel
-OpenScanHub ${sub} devel configuration
-EOF
-done)
+%package hub-conf-devel
+Summary: OpenScanHub hub devel configuration
+Provides: osh-hub-conf = %{version}-%{release}
+Conflicts: osh-hub-conf
+Requires: httpd-filesystem
+Requires: osh-hub
```
kdudka, line 115: We should explain in the commit message why we are adding the dependency.  I do not think it is obvious without the context.

osh/hub/osh-hub-httpd.conf
```diff
@@ -18,23 +18,20 @@ WSGISocketPrefix /var/run/wsgi
     SSLCertificateKeyFile   /etc/httpd/conf/localhost.key
 
     # WSGI handler
-    WSGIDaemonProcess covscan display-name=%{GROUP} locale='C.UTF-8'
-    WSGIProcessGroup covscan
-    WSGIScriptAlias /osh /usr/lib/python3.6/site-packages/osh/hub/osh-hub.wsgi process-group=covscan
+    WSGIDaemonProcess osh display-name=%{GROUP} locale='C.UTF-8'
+    WSGIProcessGroup osh
+    WSGIScriptAlias /osh @PYTHON3_SITELIB@/osh/hub/osh-hub.wsgi process-group=osh
```
kdudka, line 23: Should we rename the file to `osh-hub-httpd.conf.in` to make it clear that it is a template and not something that will work out of the box?

Alternatively, we could keep `/usr/lib/python3.6` as it is and substitute this string in the spec file.

osh.spec
lzaoral: Addressed in 4be7656 and db72d39.

osh/hub/osh-hub-httpd.conf
lzaoral: This still needs to be addressed.

lzaoral: Hmm, I'm not entirely sure if a file can be easily renamed using `setup.py` (to drop the `.in` suffix).  @kdudka,  would it be ok if the file was installed manually in the `%install` section instead?

kdudka: Yes, we can drop the change of `setup.py` and take the template file directly from the build directory in `%install`.

lzaoral: Addressed in 3e572db and f33ed48.

osh.spec
```diff
@@ -133,6 +139,9 @@ done)
 # collect static files from Django itself
 PYTHONPATH=. osh/hub/manage.py collectstatic --noinput
 
+# set path to python sitelib in the example httpd config
+sed -i 's|@PYTHON3_SITELIB@|%{python3_sitelib}|' osh/hub/osh-hub-httpd.conf.in
```
kdudka, line 143: After the in-place edit, the template file is not a template any more.  It would be cleaner to use a separate file for the non-template configuration file:
```
sed 's|@PYTHON3_SITELIB@|%{python3_sitelib}|' osh/hub/osh-hub-httpd.conf.in > osh/hub/osh-hub-httpd.conf
```

kdudka (approved):

osh.spec
lzaoral: You're correct, I was just being a bit lazy today. 😅

kdudka (approved):

PR #141, opened 2023-09-23
worker: use the `-t` option for cspodman, too

kdudka:
cspodman recently introduced the `-t` option compatible with csmock. So we can propagate the list of enabled analyzers for both the tools.

Related: https://issues.redhat.com/browse/OSH-313


jperezdealgaba (approved):
/lgtm

kdudka:
@jperezdealgaba Thanks for review!

Issue #134, opened 2023-09-06
git: update `.git-blame-ignore-revs`


lzaoral:
While the whole codebase seems to be reformatted properly now, all commits making such changes since February 2023 are not included in the `.git-blame-ignore-revs` file.  This issue tracks the progress to fix that.

PR #142, opened 2023-09-25
git: update `.git-blame-ignore-revs`

lzaoral:
Fixes: https://github.com/openscanhub/openscanhub/issues/134


kdudka (approved):
Looks good.  Thanks!

PR #145, opened 2023-10-03
hub: Disable `--use-host-cppcheck` in default profile

siteshwar:
This option should only be used if `cppcheck` is statically linked.


kdudka (changes requested):
You need to remove it from the `errata` profile, too, as long as its description contains the `(currently same as the default profile)` suffix.

kdudka (approved):

PR #148, opened 2023-10-09
ci: add CentOS Stream 9 and Fedora Latest Testing Farm jobs

lzaoral:


.packit.yaml
```diff
@@ -42,7 +42,9 @@ jobs:
       job: tests
       trigger: pull_request
       targets:
-        - epel-8-x86_64
+          - fedora-rawhide-x86_64
```
siteshwar, line 45: Fedora rawhide could be easily broken. It might be a better idea to use last stable version of fedora as the target here.

siteshwar (changes requested):

.packit.yaml
lzaoral: Yeah, you're right.  Addressed in 541cb51.

siteshwar (approved):

Issue #100, opened 2023-08-21
hub: JavaScript TypeError on waiving details page for unfinished scans


lzaoral:
1. Schedule an `ErrataDiffBuild` task.
2. Immediately go to the waiving details page for the associated scan.
3. The console in you browser will show the following error:
```js
Uncaught TypeError: document.getElementById(...) is null
    onload http://0.0.0.0:8000/waiving/1/:238
    EventHandlerNonNull* http://0.0.0.0:8000/waiving/1/:203
    [1:238:12](http://0.0.0.0:8000/waiving/1/)
    onload http://0.0.0.0:8000/waiving/1/:238
    (Async: EventHandlerNonNull)
    <anonymous> http://0.0.0.0:8000/waiving/1/:203
```

PR #163, opened 2023-10-20
hub: do not assume that the `toggle_trace_all` element is always present

lzaoral:
... because such assumption does not hold for failed or otherwise incomplete scans.

Fixes the following exception:
```js
Uncaught TypeError: document.getElementById(...) is null
    onload http://0.0.0.0:8000/waiving/1/:238
    EventHandlerNonNull* http://0.0.0.0:8000/waiving/1/:203
    [1:238:12](http://0.0.0.0:8000/waiving/1/)
    onload http://0.0.0.0:8000/waiving/1/:238
    (Async: EventHandlerNonNull)
    <anonymous> http://0.0.0.0:8000/waiving/1/:203
```

Resolves: https://github.com/openscanhub/openscanhub/issues/100


kdudka (approved):
Looks good.  Thanks!

PR #164, opened 2023-10-20
worker: use the `-t` option for `cspodman --no-scan`, too

kdudka:
As pointed out by Lukáš Zaoral, the change merged as https://github.com/openscanhub/openscanhub/pull/141 was incomplete. There is another copy of the code in `no_scan()`, which needs to be updated, too.

Fixes: commit 8bfae2dfa6410105afdfd603d0f9cb939f6ffa21
Related: https://issues.redhat.com/browse/OSH-313


jperezdealgaba (approved):
/lgtm

Issue #109, opened 2023-08-24
hub: exported XML-RPC methods contain unexpected entries


lzaoral:
We should check and mark explicitly methods and types that should be exported over XML-RPC.  For example, the http://0.0.0.0:8000/xmlrpc/worker/ page lists the following types (among others) that are exported but shouldn't be:
```python
...
worker.AnalyzerVersion
worker.AppSettings
worker.BaseNotValidException
worker.FileUpload
worker.Scan
worker.ScanBinding
worker.ScanningSession
worker.Task
worker.TaskResultsProcessor
...
```

lzaoral:
This can be in theory fixed by using a similar approach as in https://github.com/openscanhub/openscanhub/commit/fc8389f0a1a8e5da74101cc7f4c79831e59f299d.

PR #167, opened 2023-10-31
hub: do not export unwanted XML-RPC methods

lzaoral:
Resolves: #109

### DIFF
--- a/osh/hub/osh_xmlrpc/worker.py
+++ b/osh/hub/osh_xmlrpc/worker.py
@@ -27,6 +27,28 @@ from osh.hub.waiving.results_loader import TaskResultsProcessor
 
 logger = logging.getLogger(__name__)
 
+# DO NOT REMOVE!  The __all__ list contains all publicly exported XML-RPC
+# methods from this module.
+__all__ = [
+    'cancel_task',
+    'create_sb',
+    'email_scan_notification',
+    'email_task_notification',
+    'ensure_base_is_scanned_properly',
+    'ensure_cache',
+    'fail_scan',
+    'fail_task',
+    'finish_analyzers_version_retrieval',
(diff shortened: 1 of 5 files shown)


osh/hub/osh_xmlrpc/worker.py
```diff
@@ -27,6 +27,28 @@
...
+    'fail_task',
+    'finish_analyzers_version_retrieval',
+    'finish_scan',
+    'finish_task',
+    'get_scanning_args',
+    'get_su_user',
+    'interrupt_tasks',
+    'move_upload',
+    'set_scan_to_basescanning',
+    'set_scan_to_scanning',
```
rhyw, line 49: I confirmed that some of the methods are indeed gone with this patch, e.g., the `worker.AnalyzerVersion`. However I still see some of inconsistencies, i.e., I see around 34 "worker.xxx" methods under `/xmlrpc/worker/` while here I see only 17.

For example I see `worker.create_subtask`/`worker.assign_task` etc.

lzaoral: The default set of XML-RPC methods is exported by `kobo`: https://github.com/release-engineering/kobo/blob/master/kobo/hub/xmlrpc

OSH either adds new methods or overrides the ones from `kobo` to add some extra functionality.  The set of all exported Python modules is defined here: https://github.com/openscanhub/openscanhub/blob/ac9459c863ecc62bcd135f7db18b56a311c0c020/osh/hub/settings.py#L152-L178

rhyw: aha I see, thanks

rhyw (approved):

lzaoral:
Thank you for the review, @rhyw!

PR #107, opened 2023-08-22
hub: add a migration to recompute the subtask_count field

lzaoral:
The execution time on the staging instance database snapshot is around 3 minutes.

Resolves: https://github.com/openscanhub/openscanhub/issues/20

PR #168, opened 2023-10-31
hub: remove a separate model used for a single permission

lzaoral:
The `errata_xmlrpc_scan` user permission was moved to the more appropriate `Scan` model.


lzaoral:
I've recreated the migration to resolve a conflict with another migration from #107.

kdudka (approved):
Works as expected.  Thanks!

kdudka:
@lzaoral I am not able to update user permissions through web UI for a user name that contains `/`:
```
Enter a valid username. This value may contain only letters, numbers and @/./+/-/_ characters.
```

In this case, the user is a Kerberos service principal.  However, last week I hit a similar problem with a user named `worker/...`, which OSH/kobo automatically creates for each worker.  Any idea if we could somehow relax this check at `/osh/admin/kobo_auth/user/`?

Apparently, such users exist in OSH and anything but Django Admin is able to work with them.

lzaoral:
> @lzaoral I am not able to update user permissions through web UI for a user name that contains `/`:
> 
> ```
> Enter a valid username. This value may contain only letters, numbers and @/./+/-/_ characters.
> ```
> 
> In this case, the user is a Kerberos service principal. However, last week I hit a similar problem with a user named `worker/...`, which OSH/kobo automatically creates for each worker. Any idea if we could somehow relax this check at `/osh/admin/kobo_auth/user/`?
> 
> Apparently, such users exist in OSH and anything but Django Admin is able to work with them.

This is an issue unrelated to this PR.  Moved to https://github.com/openscanhub/openscanhub/issues/185.

PR #171, opened 2023-11-03
hub: remove severity field from the Checker model

lzaoral:
... because it was never actually used in this project since it was added.

Fixes: 344d530911d86c3b2d68e67c6fdfd46ce40a602a ("added south support, BZ creation support, new options for CLI")


kdudka (approved):
Looks good.  Thanks!

Issue #45, opened 2023-07-11
packaging: use a PEP 440 compliant version string


lzaoral:
Right now, `setuptools` issues the following deprecation warning:

```
/tmp/build-env-ep_0a0rc/lib/python3.11/site-packages/setuptools/dist.py:510: SetuptoolsDeprecationWarning: Invalid version: '0.9.3.git.20230710.144836.0797932'.
!!

        ********************************************************************************
        The version specified is not a valid version according to PEP 440.
        This may not work as expected with newer versions of
        setuptools, pip, and PyPI.

        By 2023-Sep-26, you need to update your project and remove deprecated calls
        or your builds will no longer be supported.

        See https://peps.python.org/pep-0440/ for details.
        ********************************************************************************

!!
```

This issue should be fixed sooner than later because as it says itself:
>  By 2023-Sep-26, you need to update your project and remove deprecated calls or your builds will no longer be supported.

kdudka:
@frenzymaddness Is there any project that uses `setuptools` to create snapshots from git in a way that version strings for commit objects created/amended later are seen by `dnf` as updates of previously installed snapshots?

frenzymadness:
As mentioned in the PEP: `Many build tools integrate with distributed version control systems like Git and Mercurial in order to add an identifying hash to the version identifier. As hashes cannot be ordered reliably such versions are not permitted in the public version field.`

So, commit hashes are not allowed in versions of Python packages. What we can do is to use dev releases like `X.Y.Z.devN` where N is a number of commits since the last release. But that works only if the history is stable so only for the main branch I guess.

Or, if we don't need the version string to be the same for the Python package and the RPM package, we can add a date and git commit hash only into RPM, see https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/#_complex_versioning_with_a_pre_and_post_release_snapshots 

Thanks to the date part of the snapshot after the version itself, the string is sortable. The logic doing that in setup.py can be extracted to a script that would produce the version we can then prepend to the specfile so only the RPMs are versioned with the snapshot.

Is this something you'd like me to take a look at? I cannot promise it in the next two weeks but it should not be a problem.

kdudka:
[...]

PR #172, opened 2023-11-06
packaging: use snapshot versioning compatible with csutils

kdudka:
... for RPM packages and `setuptools_scm` for versioning of Python modules and distribution tarballs.

Resolves: https://github.com/openscanhub/openscanhub/issues/45


frenzymadness (commented):
I like the Python part. The new shell script is kinda mysterious for me but I understand its purpose. There is only one possible confusing part - when I pull this PR and build an RPM from it, the SRPM is `osh-0.9.5.20231106.215412.g9e3f9da.snapshot_releases-1.fc38.src.rpm` but the Python sdist inside is `osh-0.9.6.dev33+g9e3f9da.tar.gz` so they both have the same commit but different versions.

kdudka:
@frenzymadness Thanks for review!  I thought we agreed on this on the last meeting in September (sorry for the delay).  The version string that we use for RPM is optimized for DNF whereas the sdist version is compatible with PEP 440.  The latter does not allow us to use `{commit_date}.{commit_time}` in the version string.

The only way to make these two version strings identical would be to rename the distribution tarball (and the top-level directory in the distribution tarball) after it is created by `setup.py sdist`.  Would this be an improvement over status quo?

kdudka:
> The only way to make these two version strings identical would be to rename the distribution tarball (and the top-level directory in the distribution tarball) after it is created by setup.py sdist.

I have implemented this in a fix-up commit: 2ae6d01e1d9f1b25f80ebae4d72db8067614babd

lzaoral (approved):
LGTM!  Thanks a lot @kdudka!  Please, also remove the `__init__.py` file from the `scrips` directory because  it is useless with these changes.

kdudka:
@lzaoral ` __init__.py` removed in commit 7d6e8b6335e4deade8b052ba449af647620ab27d.  Thanks for review!

PR #173, opened 2023-11-06
worker: do not write to the log twice

kdudka:
As reported by Lukáš Zaoral, the default installation of OSH worker was writing each log message twice by mistake.  The log `LOG_FILE` field in the `worker.conf` configuration file works as expected in the foreground mode, too.  I must have done something wrong while testing it initially.

Fixes: commit 30fadad7945a4c1de316956a77641b93a99bd649
Closes: https://github.com/openscanhub/openscanhub/pull/173


lzaoral (approved):
LGTM! Anything written to `stderr` will still be logged to journal, correct?

kdudka:
> Anything written to stderr will still be logged to journal, correct?

Yes.  Thanks for review!

PR #175, opened 2023-11-07
spec: change osh configuration packages to weak dependencies

siteshwar:
As we are in the process of containerizing hub and worker for production deployments, avoid having a hard dependency on osh configuration packages. The files provided by the configuration packages should be provided after the container build phase.


kdudka (approved):
Looks good.  Please rebase to fix the CI.

PR #81, opened 2023-07-28
containers: get rid of user creation through shell

rhyw:
Note the similar user creation hack is also in `setup_db.sh`, with slight different password setup. Since it's used in ci, I didn't update there in this PR.

PR #181, opened 2023-11-13
ci: load users from fixture instead of creating on the fly

rhyw:


scripts/setup_db.sh
```diff
@@ -6,12 +6,8 @@ EOF
...
 python3 /usr/lib/python3.*/site-packages/osh/hub/manage.py migrate
 
-python3 /usr/lib/python3.*/site-packages/osh/hub/manage.py shell <<EOF
-from django.contrib.auth import get_user_model
-User = get_user_model()
-User.objects.create_user('user', 'user@redhat.com', 'xxxxxx')
-User.objects.create_superuser('admin', 'user@redhat.com', 'velryba')
-EOF
+# create ci users
+python3 /usr/lib/python3.*/site-packages/osh/hub/manage.py loaddata fixtures/users.json
```
lzaoral, line 10: There is no need to create separate fixtures for tests in Testing Farm.  You can simply reuse those you have created in #81:
```suggestion
python3 /usr/lib/python3.*/site-packages/osh/hub/manage.py loaddata osh/hub/other/test_fixtures/users.json
```
Yes, the admin password is different (`xxxxxx` vs. `velryba`) but it is not used in those tests at all.

lzaoral (changes requested):

scripts/setup_db.sh
rhyw: Thanks, that would be one less fixture file to worry about. Addressed in latest commit.

lzaoral:
Testing Farm had an outage over the weekend.  Rescheduling:

/packit retest-failed

scripts/setup_db.sh
lzaoral: Thanks!  Unfortunately, the current change in d28a9ff does not work and the tests fail:
```shell
+ python3 /usr/lib/python3.6/site-packages/osh/hub/manage.py loaddata '/usr/lib/python3.*/site-packages/osh/hub/other/test_fixtures/users.json'
CommandError: No fixture named 'users' found.
```

The `users.json` fixture is not part of any RPM package built in COPR because it makes sense only for upstream development and, thus, won't be present in the `/usr/lib/python3.*/site-packages` directory.  Please, use the `osh/hub/other/test_fixtures/users.json` path instead which is relative to the Git repository root in which the Testing Farm tests are executed as suggested in my previous comment: https://github.com/openscanhub/openscanhub/pull/181#discussion_r1394204902

lzaoral (approved):
Perfect!  Thanks for the update, @rhyw!

rhyw:
@lzaoral thanks for suggestions on resolving pipeline failures!

Issue #110, opened 2023-08-28
hub: `scan_order`, `scans_count` and button targets in `waiving/result.html` may be incorrect


lzaoral:
Schedule three errata scans:
1. First with `NEW_PACKAGE` as the base.
2. Immediately schedule the second one  with a proper package as the baseline instead and let it finish.
3. Finally, schedule the last one with a with `NEW_PACKAGE` as the base and let it finish as well.

* http://0.0.0.0:8000/waiving/1/ will show incorrect `scan_order` value:
![image](https://github.com/openscanhub/openscanhub/assets/48823770/5a9c990a-c9d2-4d66-9f9d-b23ebac0e169)

* http://0.0.0.0:8000/waiving/2/ will show incorrect `scan_order` value (or correct if you only care about successful scans) and correct `scan_count` (or incorrect if you only care about successful scans): <img width="1128" alt="image" src="https://github.com/openscanhub/openscanhub/assets/48823770/859fd75b-2bb8-4aa5-a4b7-e93fb5127c4e">
The `Next` button will go directly to http://0.0.0.0:8000/waiving/4/ (scan number 3).

* http://0.0.0.0:8000/waiving/3/ is a base scan and it does not show `Previous` and `Next` buttons: <img width="1125" alt="image" src="https://github.com/openscanhub/openscanhub/assets/48823770/fd9b995a-3fd7-4537-af5e-077e77887bf5">

* http://0.0.0.0:8000/waiving/4/ will show incorrect `scan_order` value (or correct if you only care about successful scans) and correct `scan_count` (or incorrect if you only care about successful scans): <img width="1125" alt="image" src="https://github.com/openscanhub/openscanhub/assets/48823770/725930b8-51f5-428a-b4e9-39a43fb92f7b">
The `Previous` button will go directly to http://0.0.0.0:8000/waiving/2/ (scan number 1).

--------

The feature is definitely broken but I'm not sure what the expected behaviour should be.  Do we want to list failed scans, do we want to list base scans, ...?

lzaoral:
I suggest to modify the `scan_order` and `scan_count` variables to only consider scans scheduled by the Errata Tool.  Thus, base scans will not be taken into account because the `Previous` and `Next` buttons do not point to such scans anyway.

However, I'm not sure how to clearly show that a given scan is not included in that set and a plain `#` for the order is rather confusing.  @kdudka Do you have any ideas?

kdudka:
[...]

PR #182, opened 2023-11-13
hub: fix waiver scan navigation

lzaoral:
Resolves: https://github.com/openscanhub/openscanhub/issues/110

### DIFF
--- a/osh/hub/templates/waiving/result.html
+++ b/osh/hub/templates/waiving/result.html
@@ -150,6 +150,7 @@
 
 {% endif %}
 
+{% if not sb.scan.is_errata_base_scan %}
 <div id="result_links">
     {% if first_sb %}
         <span class="link">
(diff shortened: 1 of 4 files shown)


osh/hub/templates/waiving/result.html
```diff
@@ -32,6 +32,7 @@ <h1>{{ compare_title }}</h1>
         {% if sb.scan.tag %}
             <span id="scan_release" class="info_box black_font light_grey_bg margin_right">{{ sb.scan.tag.release.product }} release {{ sb.scan.tag.release.release }}</span>
         {% endif %}
+        <span class="info_box black_font margin_right">{{ scan_type }}</span>
```
kdudka, line 35: I would show the scan type only if it is not `Regular` to avoid confusing users in the most common case.

kdudka (approved):
Works as expected.  Thanks!

osh/hub/templates/waiving/result.html
lzaoral: Addressed in 9c2211cd78a6e7a067c873b62d247911cd74fd22.

PR #183, opened 2023-11-14
client: add --state option in `osh-cli find-tasks`

rhyw:
This PR adds ability in `osh-client find-tasks` to filter tasks by state, used in conjunction with other options. Example:
```
$ osh-cli find-tasks --help
Usage: osh-cli find-tasks [options] <query_string>

Options:
  -h, --help                  show this help message and exit
  --username=USERNAME         specify user
  --password=PASSWORD         specify password
  --hub=HUB                   specify URL of XML-RPC interface on hub
  -l, --latest                display only latest task
  -r, --regex                 query by regular expression (python, module: re)
  -p, --package               query by package name
  -n, --nvr                   query by NVR (default one)
  -s STATES, --states=STATES  query by task state. This option is used in
                              conjunction with -r, -p, or -n. Specify multiple
                              states by using it multiple times, like '-s
                              FAILED -s CLOSED'. Valid choices include FREE,
                              ASSIGNED, OPEN, CLOSED, CANCELED, FAILED,
                              INTERRUPTED, TIMEOUT, CREATED.
$ osh-cli find-tasks -r bash
3
2
1
$ osh-cli find-tasks -r bash -s CLOSED -s failed
3
```

Resolves: https://issues.redhat.com/browse/OSH-376

### DIFF
--- a/osh/client/commands/cmd_find_tasks.py
+++ b/osh/client/commands/cmd_find_tasks.py
@@ -56,9 +56,18 @@ these options: --regex, --package, --nvr"
             type="string",
             nargs=1,
             help=(f"query by task state. This option is used in conjunction with -r, -p, or -n. "
-                  f"Specify multiple states by using it multiple times, like '-s failed -s closed'. "
-                  f"Valid choices include {', '.join([s.lower() for s in TASK_STATES])}.")
+                  f"Specify multiple states by using it multiple times, like '-s FAILED -s CLOSED'. "
+                  f"Valid choices include {', '.join(TASK_STATES)}.")
         )
+
+    def _validate_states(self, states):
+        """
+        State validator, raise error in case user specifies invalid task states
+        """
+        valid_states = list(TASK_STATES)
+        invalid_states = [s for s in states if s.upper() not in valid_states]
+        if invalid_states:
+            self.parser.error(f"Invalid state(s) specified: {', '.join(invalid_states)}.")
 
     def run(self, *args, **kwargs):
         regex = kwargs.pop("regex")
@@ -85,6 +94,7 @@ these options: --regex, --package, --nvr"
             # nvr is default one, so we don't care if it's specified
             query['nvr'] = query_string
         if states:
+            self._validate_states(states)
             query['states'] = [TASK_STATES[state.upper()] for state in states]
         task_ids = self.hub.scan.find_tasks(query)
 
--- a/osh/client/commands/cmd_find_tasks.py
+++ b/osh/client/commands/cmd_find_tasks.py
@@ -3,6 +3,8 @@
 
 import sys
 
+from kobo.client.constants import TASK_STATES
+
 import osh.client
 
 
@@ -47,10 +49,21 @@ these options: --regex, --package, --nvr"
             action="store_true",
             help="query by NVR (default one)"
         )
+        self.parser.add_option(
+            "-s",
+            "--states",
+            action="append",
+            type="string",
+            nargs=1,
+            help=(f"query by task state. This option is used in conjunction with -r, -p, or -n. "
+                  f"Specify multiple states by using it multiple times, like '-s failed -s closed'. "
+                  f"Valid choices include {', '.join([s.lower() for s in TASK_STATES])}.")
+        )
 
     def run(self, *args, **kwargs):
         regex = kwargs.pop("regex")
         package_name = kwargs.pop("package")
+        states = kwargs.get("states")
 
         latest = kwargs.pop("latest")
 
@@ -71,6 +84,8 @@ these options: --regex, --package, --nvr"
         else:
             # nvr is default one, so we don't care if it's specified
             query['nvr'] = query_string
+        if states:
+            query['states'] = [TASK_STATES[state.upper()] for state in states]
         task_ids = self.hub.scan.find_tasks(query)
 
         if not task_ids:
--- a/osh/client/completion/_osh-cli
+++ b/osh/client/completion/_osh-cli
@@ -76,8 +76,7 @@ _osh-cli_args()
                 '(-r --regex)'{-r,--regex}'[query by regular expression (python, module: re)]'
                 '(-p --pacakge)'{-p,--package}'[query by package name]'
                 '(-n --nvr)'{-n,--nvr}'[query by NVR (default one)]'
-                '(-s --state)'{-s,--state}'[query by task state. This option is used \
-in conjunction with -r, -p, or -n.]'
+                '*'{-s+,--state=}'[query by task state. This option is used in conjunction with -r, -p, or -n.]:osh:->states'
             )
             ;;
 
@@ -171,6 +170,10 @@ in conjunction with -r, -p, or -n.]'
 
         profiles)
             _osh-cli_complete profiles
+            ;;
+
+        states)
+            _osh-cli_complete task-states
             ;;
 
         warn)
--- a/osh/client/completion/_osh-cli
+++ b/osh/client/completion/_osh-cli
@@ -76,6 +76,8 @@ _osh-cli_args()
                 '(-r --regex)'{-r,--regex}'[query by regular expression (python, module: re)]'
                 '(-p --pacakge)'{-p,--package}'[query by package name]'
                 '(-n --nvr)'{-n,--nvr}'[query by NVR (default one)]'
+                '(-s --state)'{-s,--state}'[query by task state. This option is used \
+in conjunction with -r, -p, or -n.]'
             )
             ;;
 
--- a/osh/client/completion/osh-cli.bash
+++ b/osh/client/completion/osh-cli.bash
@@ -69,6 +69,7 @@ _osh_cli()
         COMPREPLY+=( $(compgen -W '-r --regex  ' -- "${cur}") )
         COMPREPLY+=( $(compgen -W '-p --package' -- "${cur}") )
         COMPREPLY+=( $(compgen -W '-n --nvr    ' -- "${cur}") )
+        COMPREPLY+=( $(compgen -W '-s --state  ' -- "${cur}") )
         ;;
 
     list-tasks)
--- a/osh/client/completion/osh-cli.bash
+++ b/osh/client/completion/osh-cli.bash
@@ -65,11 +65,19 @@ _osh_cli()
         ;;
 
     find-tasks)
-        COMPREPLY+=( $(compgen -W '-l --latest ' -- "${cur}") )
-        COMPREPLY+=( $(compgen -W '-r --regex  ' -- "${cur}") )
-        COMPREPLY+=( $(compgen -W '-p --package' -- "${cur}") )
-        COMPREPLY+=( $(compgen -W '-n --nvr    ' -- "${cur}") )
-        COMPREPLY+=( $(compgen -W '-s --state  ' -- "${cur}") )
+        case "${prev}" in
+        -s|--state)
+            _osh_cli_complete task-states
+            return
+            ;;
+        *)
+            COMPREPLY+=( $(compgen -W '-l --latest ' -- "${cur}") )
+            COMPREPLY+=( $(compgen -W '-r --regex  ' -- "${cur}") )
+            COMPREPLY+=( $(compgen -W '-p --package' -- "${cur}") )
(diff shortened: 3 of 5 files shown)


lzaoral (changes requested):
This is a very useful option, thanks @rhyw!

Additional comments:
* You have added the `-s`/`--state` completions for `zsh` but OSH also has completions for `bash`: https://github.com/openscanhub/openscanhub/blob/main/osh/client/completion/osh-cli.bash
* Would it be possible to select multiple states at the same time? For example if someone would like to select both `failed` and `interrupted` tasks.

edit; typo

rhyw:
seleting multiple values is beyond the scope of the original request. but I agree it would be a bonus.

rhyw:
The legacy `optparse` doesn't support multiple options with `nargs` and `choices` very well. I was blocked for some time till I found a work around - the validation was done manually.

I've fixed the issues per review. PTAL.

rhyw:
@lzaoral any clues of the shell check complaints?

```
Error: SHELLCHECK_WARNING:
./osh/client/completion/osh-cli.bash:218:30: warning[SC2207]: Prefer mapfile or read -a to split command output (or quote to avoid splitting).
#  215|               *)
#  216|                   COMPREPLY+=( $(compgen -W "--base-config=" -- "${cur}") )
#  217|                   COMPREPLY+=( $(compgen -W "--base-nvr=" -- "${cur}") )
#  218|->                 COMPREPLY+=( $(compgen -W "--srpm= --base-srpm=" -- "${cur}") )
#  219|                   ;;
#  220|               esac
#  221|               ;;

✅ SARIF report was successfully uploaded to GitHub
```

The `+=` seem to be used in various places in this file.

lzaoral:
@rhyw ShellCheck is not complaining about the `+=` operator but about the fact that the string that `$(compgen -W "--srpm= --base-srpm=" -- "${cur}")` expands to will be split using `IFS` which is by default a space.  In this case, the warning is harmless because that's exactly what we want.  Each space-separated part of the reply is a one possible completion candidate.

If you want to, you can silence this particular warning (`SC2207`) for the whole file because it affects all instances of `compgen` like this:
https://github.com/koalaman/shellcheck/wiki/Ignore#ignoring-all-instances-in-a-file-044

rhyw:
Addressed in b9e25e920e952849ec220e1efa654aeeabc365c5

osh/client/commands/cmd_find_tasks.py
```diff
@@ -71,6 +93,8 @@ def run(self, *args, **kwargs):
         else:
             # nvr is default one, so we don't care if it's specified
             query['nvr'] = query_string
+        if state_name is not None:
+            query['state'] = get_state_from_name(state_name)
```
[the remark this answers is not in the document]
lzaoral: No, there is no need to update that line of code because those two expressions in https://github.com/openscanhub/openscanhub/pull/183#discussion_r1394148929 are equivalent.  The `__get__` magic method of kobo's `Enum` class returns the ID of the state that is stored in the database.  Moreover, it is also more efficient with respect to the size of the `TASK_STATES` enum.

osh/client/completion/_osh-cli
```diff
@@ -76,6 +76,8 @@ _osh-cli_args()
                 '(-r --regex)'{-r,--regex}'[query by regular expression (python, module: re)]'
                 '(-p --pacakge)'{-p,--package}'[query by package name]'
                 '(-n --nvr)'{-n,--nvr}'[query by NVR (default one)]'
+                '(-s --state)'{-s,--state}'[query by task state. This option is used \
```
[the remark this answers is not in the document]
lzaoral: Sorry, I overlooked this in the previous review round.  The `-s` option takes an argument so the completions should reflect that as well.  I'll post a patch, hopefully this week, that also implements the automated completions of task states themselves so that you don't have to learn the magic behind `zsh` and `bash` completions.

ashwindasr:
Thanks @rhyw for working on this and @lzaoral for the reviews. Just checking in to see if there is anything else this PR is waiting on or will it be scheduled to merge soon?

rhyw:
@ashwindasr I think I've addressed the issues, but it still needs to be approved before moving forward.

lzaoral (commented):
I'll wait for @kududka to finish his review.  After that I'll poist a patch with shell completions.

rhyw:
Updated to behave with uppercase by default, and code slightly revised to be case insensitive. Example:

```
$ docker exec -it osh-client python3 osh/client/osh-cli find-tasks --help
Usage: osh-cli find-tasks [options] <query_string>

Options:
  -h, --help                  show this help message and exit
  --username=USERNAME         specify user
  --password=PASSWORD         specify password
  --hub=HUB                   specify URL of XML-RPC interface on hub
  -l, --latest                display only latest task
  -r, --regex                 query by regular expression (python, module: re)
  -p, --package               query by package name
  -n, --nvr                   query by NVR (default one)
  -s STATES, --states=STATES  query by task state. This option is used in
                              conjunction with -r, -p, or -n. Specify multiple
                              states by using it multiple times, like '-s
                              FAILED -s CLOSED'. Valid choices include FREE,
                              ASSIGNED, OPEN, CLOSED, CANCELED, FAILED,
                              INTERRUPTED, TIMEOUT, CREATED.

without '-l' option, newest task is at the beginning of a list, unfinished
tasks are at the end; you should pick one of these options: --regex,
--package, --nvr

$ osh-cli find-tasks -r firefox-102.11 | wc -l
      34
$ osh-cli find-tasks -r firefox-102.11 -s CLOSED -s failed | wc -l
      27
$ osh-cli find-tasks -r firefox-102.11 -s CLOSED -s Failed -s INTERRUPTED -s canceled | wc -l
      34
$ osh-cli find-tasks -r firefox-102.11 -s CLOSED -s Failed -s FOO -s invalid-state
Usage: osh-cli find-tasks [options] <query_string>

osh-cli: error: Invalid state(s) specified: FOO, invalid-state.

```

rhyw:
@kdudka I guess there's some amends that I forgot to force push after local test. Should be fixed now!

osh/client/commands/cmd_find_tasks.py
```diff
@@ -47,10 +49,30 @@ def options(self):
...
+                  f"Specify multiple states by using it multiple times, like '-s FAILED -s CLOSED'. "
+                  f"Valid choices include {', '.join(s.upper() for s in TASK_STATES)}.")
+        )
+
+    def _validate_states(self, states):
+        """
+        State validator, raise error in case user specifies invalid task states
+        """
+        valid_states = [state.lower() for state in TASK_STATES]
+        invalid_states = [s for s in states if s.lower() not in valid_states]
```
lzaoral, line 68: If upper case is going to be the default, the code can be further simplified like this:
```suggestion
                  f"Valid choices include {', '.join(TASK_STATES)}.")
        )

    def _validate_states(self, states):
        """
        State validator, raise error in case user specifies invalid task states
        """
        valid_states = list(TASK_STATES)
        invalid_states = [s for s in states if s.upper() not in valid_states]
```

osh/client/completion/_osh-cli
lzaoral, line 79: Completions for ZSH:
```diff
diff --git a/osh/client/completion/_osh-cli b/osh/client/completion/_osh-cli
index 526f9af..8b77530 100644
--- a/osh/client/completion/_osh-cli
+++ b/osh/client/completion/_osh-cli
@@ -76,8 +76,7 @@ _osh-cli_args()
                 '(-r --regex)'{-r,--regex}'[query by regular expression (python, module: re)]'
                 '(-p --pacakge)'{-p,--package}'[query by package name]'
                 '(-n --nvr)'{-n,--nvr}'[query by NVR (default one)]'
-                '(-s --state)'{-s,--state}'[query by task state. This option is used \
-in conjunction with -r, -p, or -n.]'
+                '*'{-s+,--state=}'[query by task state. This option is used in conjunction with -r, -p, or -n.]:osh:->states'
             )
             ;;

@@ -172,6 +171,10 @@ in conjunction with -r, -p, or -n.]'
             _osh-cli_complete profiles
             ;;

+        states)
+            _osh-cli_complete task-states
+            ;;
+
         warn)
             local -a warn_levels
             warn_levels=(
```
It is rather tricky to write more complex completions for ZSH ao the explanation for the changes is the following:
* `*` for option means that repetition is allow, otherwise the option would be offered only once
* `(-s --state)` should not be present because it means that those two options conflict with each other.  So once `-s` was used, `--state` would not be offered.
* `-s+,--state=` means that the short and long options require an argument.
* `:osh:->states` defines the action to complete the option arguments.

osh/client/completion/osh-cli.bash
```diff
@@ -69,6 +69,7 @@ _osh_cli()
         COMPREPLY+=( $(compgen -W '-r --regex  ' -- "${cur}") )
         COMPREPLY+=( $(compgen -W '-p --package' -- "${cur}") )
         COMPREPLY+=( $(compgen -W '-n --nvr    ' -- "${cur}") )
+        COMPREPLY+=( $(compgen -W '-s --state  ' -- "${cur}") )
```
lzaoral, line 72: Completions for BASH:
```diff
diff --git a/osh/client/completion/osh-cli.bash b/osh/client/completion/osh-cli.bash
index 4795c22..964e379 100644
--- a/osh/client/completion/osh-cli.bash
+++ b/osh/client/completion/osh-cli.bash
@@ -65,11 +65,19 @@ _osh_cli()
         ;;

     find-tasks)
-        COMPREPLY+=( $(compgen -W '-l --latest ' -- "${cur}") )
-        COMPREPLY+=( $(compgen -W '-r --regex  ' -- "${cur}") )
-        COMPREPLY+=( $(compgen -W '-p --package' -- "${cur}") )
-        COMPREPLY+=( $(compgen -W '-n --nvr    ' -- "${cur}") )
-        COMPREPLY+=( $(compgen -W '-s --state  ' -- "${cur}") )
+        case "${prev}" in
+        -s|--state)
+            _osh_cli_complete task-states
+            return
+            ;;
+        *)
+            COMPREPLY+=( $(compgen -W '-l --latest ' -- "${cur}") )
+            COMPREPLY+=( $(compgen -W '-r --regex  ' -- "${cur}") )
+            COMPREPLY+=( $(compgen -W '-p --package' -- "${cur}") )
+            COMPREPLY+=( $(compgen -W '-n --nvr    ' -- "${cur}") )
+            COMPREPLY+=( $(compgen -W '-s --state= ' -- "${cur}") )
+            ;;
+        esac
         ;;

     list-tasks)

```

lzaoral (changes requested):
The following patch adds the `task-states` option to `osh/client/completions/main.py`:
```diff
diff --git a/osh/client/completion/main.py b/osh/client/completion/main.py
index edf085d..5ba5b03 100644
--- a/osh/client/completion/main.py
+++ b/osh/client/completion/main.py
@@ -7,6 +7,7 @@ import pickle
 from datetime import datetime, timedelta

 from kobo.client import HubProxy
+from kobo.client.constants import TASK_STATES

 from osh.common.conf import get_config_dict

@@ -55,6 +56,13 @@ def fetch_mock_configs(hub):
     return [x['name'] for x in hub.mock_config.all() if x['enabled']]


+def fetch_task_states(hub):
+    """
+    Return all possible task states
+    """
+    return list(TASK_STATES)
+
+
 def load_from_cache(action):
     """
     Try to load up to date data from the cache
@@ -99,7 +107,8 @@ def main(args):
 ACTIONS = {
     'analyzers': fetch_analyzers,
     'mock-configs': fetch_mock_configs,
-    'profiles': fetch_profiles
+    'profiles': fetch_profiles,
+    'task-states': fetch_task_states,
 }


```

osh/client/commands/cmd_find_tasks.py
rhyw: updated

rhyw:
bash/zsh completion seem to bring extra maintenance burden when new options are added. Thanks for your help @lzaoral . All updated according to the suggestions above.

kdudka (approved):
Works as expected, including the bash/zsh completions.

rhyw:
Thanks for the review/verification.

PR #184, opened 2023-11-15
hub: add management command to filter failed tasks impacted by OSH-362

rhyw:
Usage:

    python3 osh/hub/manage.py find_failed_tasks

### DIFF
--- a/osh/hub/scan/management/commands/find_failed_tasks.py
+++ b/osh/hub/scan/management/commands/find_failed_tasks.py
@@ -1,0 +1,45 @@
+from datetime import datetime
+
+import requests
+from django.core.management.base import BaseCommand
+from kobo.hub.models import Task
+
+start_date = datetime(2023, 10, 18)
+end_date = datetime(2023, 10, 19, 23, 59, 59)
+CSMOCK_VERSION = "csmock-3.5.0.20231018.081735.gf8bfc23.internal-1"
+
+
+def get_task_log(task_id, log_name="stdout.log"):
+    hub = 'https://cov01.lab.eng.brq2.redhat.com/osh'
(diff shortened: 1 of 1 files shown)


rhyw:
This is a one-time helper command. I assume we don't need it to be merged.

osh/hub/scan/management/commands/find_failed_tasks.py
```diff
@@ -0,0 +1,45 @@
+from datetime import datetime
+
+import requests
+from django.core.management.base import BaseCommand
+from kobo.hub.models import Task
+
+start_date = datetime(2023, 10, 18)
+end_date = datetime(2023, 10, 19, 23, 59, 59)
+CSMOCK_VERSION = "csmock-3.5.0.20231018.081735.gf8bfc23.internal-1"
```
jperezdealgaba, line 9: Although this is a one time command, wouldn't it be better to not be tied to a specific version of csmock?

lzaoral (changes requested):
As agreed with @kdudka, this feature is directly tied to your specific and private deployment of OpenScanHub and is **not applicable** to OpenScanHub in general.  Therefore, the upstream repository is not the right place for such changes.

edit: typo

kdudka:
Yes, this does not make any sense from upstream point of view.  https://issues.redhat.com/browse/OSH-362 is not publicly available and there is no description of what and why is being done by the changes.  Also the name `find_failed_tasks` is misleading when we are in fact looking for `CLOSED` tasks.  Let's follow up on the internal Jira issue.

rhyw:
I'm just finding somewhere for a review of whether the filter is doing what it's supposed to. I don't have a better idea how this could be reviewed. Never mind, I also updated in the jira.

PR #186, opened 2023-11-16
hub: fix the `All defects` links on waiver page

kdudka:
... such that they really point to all defects and not to the compatibility symlinks to important defects only.

This fixes a user experience issue introduced by the following change in csmock:
https://github.com/csutils/csmock/pull/126

### DIFF
--- a/osh/hub/waiving/views.py
+++ b/osh/hub/waiving/views.py
@@ -193,10 +193,10 @@ def add_logs_to_context(sb):
                                  (FIXED_HTML_FILE, 'HTML'),
                                  (FIXED_DIFF_FILE, 'JSON')], logs_list))
     logs.append(create_log_dict('All defects', 'Document_content_32.png',
-                                log_prefix + '.html',
-                                [(log_prefix + '.err', 'TXT'),
-                                 (log_prefix + '.html', 'HTML'),
-                                 (log_prefix + '.js', 'JSON')], logs_list))
+                                log_prefix + '-all.html',
+                                [(log_prefix + '-all.err', 'TXT'),
+                                 (log_prefix + '-all.html', 'HTML'),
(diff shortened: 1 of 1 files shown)


osh/hub/waiving/views.py
```diff
@@ -193,10 +193,10 @@ def add_logs_to_context(sb):
...
                                  (FIXED_DIFF_FILE, 'JSON')], logs_list))
     logs.append(create_log_dict('All defects', 'Document_content_32.png',
-                                log_prefix + '.html',
-                                [(log_prefix + '.err', 'TXT'),
-                                 (log_prefix + '.html', 'HTML'),
-                                 (log_prefix + '.js', 'JSON')], logs_list))
+                                log_prefix + '-all.html',
+                                [(log_prefix + '-all.err', 'TXT'),
+                                 (log_prefix + '-all.html', 'HTML'),
+                                 (log_prefix + '-all.js', 'JSON')], logs_list))
```
lzaoral, line 199: Off topic: We should probably change the to extension from `.js` to `.json` to prevent confusion.

lzaoral: (in some other PR, of course)

kdudka: You mean in `csmock`?  That could be tricky as it is part of its public interface: https://www.mankier.com/1/csmock#Output_Format

lzaoral: Yes, I didn't say it would be easy. 😄 BTW, the RFC defining JSON also specifies `.json` as the recommended extension: https://datatracker.ietf.org/doc/html/rfc8259#section-11

lzaoral: This is not critical but I've created an issue for this: https://github.com/csutils/csmock/issues/141

kdudka:
@jperezdealgaba Thanks for review!

PR #188, opened 2023-11-23
packaging: clean-up the Makefile

lzaoral:


Makefile
```diff
@@ -6,50 +6,37 @@ help:
...
 	@echo " help                    show this text"
-	@echo " build			builds basic development environment"
+	@echo " build                   builds basic development environment"
 	@echo " clean                   remove python bytecode and temp files"
 	@echo " clean-dev               also cleans up development environment"
-	@echo " cli-test		run cli sanity tests"
-	@echo " full-dev		builds full development environment"
+	@echo " deploy                  builds and starts basic development environment"
+	@echo " full-build              builds full development environment"
+	@echo " full-deploy             builds and starts basic development environment"
```
kdudka, line 14: This seems to be a copy/paste error: `basic` vs. `full`

Should we rather say what the difference between them is so that the help text provides any value?

.github/workflows/lint.yml
```diff
@@ -22,13 +22,11 @@ jobs:
...
+
           # Check if we are running CI in a fork
           if ! git remote -v | grep -q 'origin.*openscanhub/openscanhub'; then
             REPO=openscanhub
             git remote add ${REPO} https://github.com/openscanhub/openscanhub.git
+            git fetch ${REPO}
           fi
-          git fetch ${REPO} ${BRANCH}
-          pre-commit run --show-diff-on-failure --color=always --from-ref ${REPO}/${BRANCH} --to-ref HEAD
+          make lint REPO=$REPO
```
kdudka, line 32: The CI job is now red.  Could it be related to this change?
```
An unexpected error has occurred: CalledProcessError: command: ('/usr/bin/git', 'diff', '--name-only', '--no-ext-diff', '-z', '/main..HEAD')
return code: 128
stdout: (none)
stderr:
    fatal: ambiguous argument '/main..HEAD': unknown revision or path not in the working tree.
    Use '--' to separate paths from revisions, like this:
    'git <command> [<revision>...] -- [<file>...]'
Check the log at /github/home/.cache/pre-commit/pre-commit.log
make: *** [Makefile:91: lint] Error 3
Error: Process completed with exit code 2.
```

kdudka (changes requested):
Looks good.  Some remarks inline.

.github/workflows/lint.yml
lzaoral: Addressed in 97b6caf.

Makefile
lzaoral: Nice catch!  Addressed in  ddbd30ba5e5bc68e0e258a95e57f5df4dc10eb12.

PR #189, opened 2023-11-27
containers: run client under a regular user

lzaoral:


lzaoral:
I'll fix the typo in the commit message of 9123d53572f4a910156d669619014776a86ed311 when rebasing before merge.

compose.yaml
```diff
@@ -35,7 +35,7 @@ services:
       - hub-tasks:/src/osh/hub/tasks
       - hub-upload:/src/osh/hub/upload
       - hub-worker:/src/osh/hub/worker
-      - coverage:/cov
+      - coverage:/cov:U
```
kdudka, line 38: The `podman-run(1)` man page says:
> Warning use with caution since this modifies the host filesystem.

Cannot this change surprise users?

lzaoral: That should be harmless because `coverage` is a volume managed by `podman`.  It is not a mount accessible by the host (unless you do access that volume explicitly with `podman mount`).

The issue is that the user of the first container that mounts that volume internally will own that filesystem.  Therefore, if `osh-worker` is the first, the owner UID/GID won't match other containers (as that container's user is `root`).  The main motivation for `:U` is to ensure that the contests of that volume will be owned by UID `1000`.

lzaoral: The plan is to remove the `:U` attribute in #190 because after that all containers will be running under a user with the same UID/GID.

kdudka: Thank you for clarifying!  Assuming the container will work for users with different UID (and will not create files on the host file system that the user cannot access), I am fine with that.

kdudka (approved):

PR #190, opened 2023-11-27
[WIP] worker: run under regular user

lzaoral:


kdudka:
Do we want to use the `osh-worker` user for RPM-based deployment but the `osh` user for container-based deployment?

kdudka:
This is not going to be a smooth migration.  We will need to transfer existing data of the `csmock` user (such as auth tokens and download cache).  This data is used by both `csmock` and `cspodman`.  We should also check how this will affect task cancellation because `osh-worker` will not have permission to kill processes of other users any more.  This reminds me a `csmock` patch that I was experimenting with 3 years ago: https://github.com/csutils/csmock/compare/main...kdudka:csmock:script

hanchuntao:
After research, we cannot use this method to limit osh-worker.service memory, because some children process will escape from the service, so after talking with @kdudka, we revert the memory list setting at https://gitlab.cee.redhat.com/osh/cspodman/-/merge_requests/68, so close this PR

kdudka:
@hanchuntao I think you wanted to write the above comment on #193 instead.

PR #170, opened 2023-11-01
hub: remove the `ScanningSession` model

lzaoral:
Resolves: https://issues.redhat.com/browse/OSH-298

PR #191, opened 2023-11-28
worker: remove references to `ScanningSession`

lzaoral:
Note that https://github.com/openscanhub/openscanhub/pull/170 must be merged and deployed first!


lzaoral:
@kdudka The PR contains only a single commit but for some reason, the UI is broken shows commit from #170 as well.

PR #192, opened 2023-11-30
hub: remove all remaining mentions to `ScanningSession`

lzaoral:
Fixes: 460aeea953a1918e55ad2ce4891c573be322297e ("hub: make ErrataDiffBuild task independent of ScanningSession")
Related: https://github.com/openscanhub/openscanhub/pull/170
Related: https://github.com/openscanhub/openscanhub/pull/191

PR #193, opened 2023-11-30
draft: worker: set memory to 95% in the osh-worker unit file

hanchuntao:
it is safer if we kill either the wild scan task, or the osh-worker daemon itself, rather than killing core part of the operating system and let the daemon pick other scanning task to fail or misbehave in such a broken environment

Related: https://issues.redhat.com/browse/OSH-394
Closes: https://github.com/openscanhub/openscanhub/pull/193

### DIFF
--- a/osh/worker/osh-worker.service
+++ b/osh/worker/osh-worker.service
@@ -10,6 +10,8 @@ ExecStart=/usr/sbin/osh-worker -f
 KillMode=process
 Restart=on-failure
 StandardOutput=null
+MemoryHigh=95%
+MemoryMax=95%
 
 [Install]
 WantedBy=multi-user.target


siteshwar:
How is it related to https://github.com/openscanhub/openscanhub/pull/192?

hanchuntao:
> How is it related to #192?
Sorry, make a mistake, should be #193

lzaoral:
@kdudka Wouldn't it be better to use a bit lower value for `MemoryHigh=`?  According to the docs for `MemoryHigh=` [1]:

> Specify the throttling limit on memory usage of the executed processes in this unit. Memory usage may go above the limit if unavoidable, but the processes are heavily slowed down and memory is taken away aggressively in such cases. This is the main mechanism to control memory usage of a unit.

Docs for `MemoryMax=` suggest that as well:
> It is recommended to use MemoryHigh= as the main control mechanism and use MemoryMax= as the last line of defense.

[1] https://www.freedesktop.org/software/systemd/man/latest/systemd.resource-control.html

kdudka:
@lzaoral I am not sure.  If we slow down the whole process tree under `osh-worker`, there will be nothing else on the OSH worker machine that could use the remaining memory for anything useful before the process tree is killed.  Consequently, users will wait unnecessarily longer to get a notification that their task was interrupted.

@hanchuntao Were you able to trigger the OOM killer on a machine with a lower amount of memory to see whether it works as expected?

hanchuntao:
> @lzaoral I am not sure. If we slow down the whole process tree under `osh-worker`, there will be nothing else on the OSH worker machine that could use the remaining memory for anything useful before the process tree is killed. Consequently, users will wait unnecessarily longer to get a notification that their task was interrupted.
> 
> @hanchuntao Were you able to trigger the OOM killer on a machine with a lower amount of memory to see whether it works as expected?

sure, I will trigger and test it

hanchuntao:
**Test steps:**
1) Change the osh-worker service: /usr/lib/systemd/system/osh-worker.service
```
MemoryHigh=20%
MemoryMax=20%
```
```
systemctl daemon-reload
systemctl restart osh-worker
```

2) Check the memory status
```
[root@osh-test-worker-001 ~]# free -hl
               total        used        free      shared  buff/cache   available
Mem:            62Gi       2.3Gi        53Gi       264Mi       7.2Gi        59Gi
Low:            62Gi       8.6Gi        53Gi
High:             0B          0B          0B
Swap:           63Gi          0B        63Gi
[root@osh-test-worker-001 ~]#
```

3) Create memory intensive process
`stress --vm 10 --vm-bytes 10G --timeout 300s`

4) Check the memory again
```
[root@osh-test-worker-001 ~]# free -hl
               total        used        free      shared  buff/cache   available
Mem:            62Gi        62Gi       382Mi        61Mi       282Mi        45Mi
Low:            62Gi        61Gi       382Mi
High:             0B          0B          0B
Swap:           63Gi        24Gi        39Gi
[root@osh-test-worker-001 ~]# 
```

5) The service is still "active", looks like not work as expect
```
[root@osh-test-worker-001 ~]# systemctl status  osh-worker
● osh-worker.service - OpenScanHub worker daemon
     Loaded: loaded (/usr/lib/systemd/system/osh-worker.service; enabled; preset: disabled)
     Active: active (running) since Tue 2023-12-05 03:42:21 UTC; 28min ago
    Process: 1555908 ExecStartPre=test -r /etc/osh/worker.conf (code=exited, status=0/SUCCESS)
   Main PID: 1555909 (osh-worker)
      Tasks: 1 (limit: 407647)
     Memory: 14.1M (max: 12.4G available: 12.4G)
        CPU: 12.979s
     CGroup: /system.slice/osh-worker.service
             └─1555909 /usr/bin/python3 /usr/sbin/osh-worker -f

Dec 05 03:42:21 osh-test-worker-001.osh-001.preprod.iad2.dc.redhat.com systemd[1]: osh-worker.service: Deactivated successfully.
Dec 05 03:42:21 osh-test-worker-001.osh-001.preprod.iad2.dc.redhat.com systemd[1]: Stopped OpenScanHub worker daemon.
Dec 05 03:42:21 osh-test-worker-001.osh-001.preprod.iad2.dc.redhat.com systemd[1]: osh-worker.service: Consumed 1.029s CPU time.
Dec 05 03:42:21 osh-test-worker-001.osh-001.preprod.iad2.dc.redhat.com systemd[1]: Starting OpenScanHub worker daemon...
Dec 05 03:42:21 osh-test-worker-001.osh-001.preprod.iad2.dc.redhat.com systemd[1]: Started OpenScanHub worker daemon.
[root@osh-test-worker-001 ~]# 
```

lzaoral:
> @lzaoral I am not sure. If we slow down the whole process tree under `osh-worker`, there will be nothing else on the OSH worker machine that could use the remaining memory for anything useful before the process tree is killed. Consequently, users will wait unnecessarily longer to get a notification that their task was interrupted.

Hmm, from the point of a user.  I'd be fine with slower task execution than with an interrupted task.  Ideally, only the hungriest task should be slowed down/killed but I'm not sure if `systemd` is able to do that.  (Also, if you insist on unconditional killing, is it even necessary to set `MemoryHigh=` in the first place?)

> The service is still "active", looks like not working as expected, I think it is caused by the swap

@hanchuntao You should take a look at the whole journal history (`journalctl -u osh-worker.service`).  The service unit file specifies `Restart=on-failure` so I'd expect it to be running again after is it killed.

kdudka:
@lzaoral As I understand it, the setting applies to the whole process group, which means all the the tasks being processed by the worker.  I do not think we can make `systemd` slow down a specific OSH task to make the other OSH tasks complete successfully on the same worker.

hanchuntao:
> @hanchuntao You should take a look at the whole journal history (`journalctl -u osh-worker.service`). The service unit file specifies `Restart=on-failure` so I'd expect it to be running again after it is killed.

@lzaoral I checked the whole journal history, and found the service did not restart

kdudka:
@hanchuntao As I understand it, you expect the top-level process to consume all the memory.  That is, however, not the scenario that happens on OSH workers.  We need to submit a few OSH tasks that are hungry of memory, such as:
```
$ osh-cli mokc-build --config cspodman --nvr odh-trustyai-service-operator-container-v1.30.0-2
```
... and check what happens when the memory limit is exceeded.

I do not think it makes sense to set `MemoryHigh` to the same value as `MemoryMax`.  I would use only `MemoryMax` unless we have an evidence that the use of `MemoryHigh` improves anything in a real-word scenario.

hanchuntao:
@kdudka yes, you are right, but to test it, I need to update the settings for all of the instances, because I don't know which instance will run this task

kdudka:
As explained on Slack, this can be easily tested on an OSH deployment with only one enabled worker.

kdudka:
It turned out that the resource limits set in the `osh-worker` unit file do not work as expected because OSH worker runs `csmock` through `su`, which makes the limits ineffective.  We should reconsider this once #190 is resolved.

Issue #200, opened 2023-12-07
hub: do not use `noarch` as the default worker architecture


lzaoral:
See `$TITLE`.

PR #204, opened 2023-12-07
hub: use native architectures by default

lzaoral:
Resolves: https://github.com/openscanhub/openscanhub/issues/200

Blocked by: https://github.com/release-engineering/kobo/pull/237

### DIFF
--- a/osh/worker/tasks/task_build.py
+++ b/osh/worker/tasks/task_build.py
@@ -18,7 +18,7 @@ class OSHTaskBase(TaskBase):
     enabled = False
 
     # list of supported architectures
-    arches = ["noarch"]
+    arches = ["noarch", "aarch64", "x86_64"]
 
     # list of channels
     channels = ["default"]
--- a/osh/worker/tasks/task_build.py
+++ b/osh/worker/tasks/task_build.py
@@ -10,15 +10,32 @@ from kobo.worker import TaskBase
 from osh.worker.csmock_runner import CsmockRunner
 
 
-class Build:
-    enabled = True
+class OSHTaskBase(TaskBase):
+    """
+    Default task settings
+    """
+    # each final class should set this to true
+    enabled = False
 
-    arches = ["noarch"]     # list of supported architectures
-    channels = ["default"]  # list of channels
-    exclusive = False       # leave False here unless you really know what you're doing
-    foreground = False      # if True the task is not forked and runs in the worker process (no matter you run worker without -f)
+    # list of supported architectures
+    arches = ["noarch"]
+
+    # list of channels
+    channels = ["default"]
+
+    # exclusive tasks have the highest possible priority
+    # leave False unless you really know what you're doing
+    exclusive = False
+
+    # if True the task is not forked and runs in the worker process
+    # (no matter you run worker without -f)
+    foreground = False
+
+    # determines how many resources is used when processing the task
     weight = 1.0
 
+
+class Build(OSHTaskBase):
     def run(self):
         mock_config = self.args.pop("mock_config")
         build = self.args.pop("build", {})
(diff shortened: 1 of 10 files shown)


scripts/add-worker-arch.py
```diff
@@ -0,0 +1,28 @@
...
+# SPDX-FileCopyrightText: Copyright contributors to the OpenScanHub project.
+
+import os
+import platform
+
+os.environ['DJANGO_SETTINGS_MODULE'] = 'osh.hub.settings'
+
+
+def main():
+    from kobo.hub.models import Arch, Worker  # noqa:  E402
```
lzaoral, line 12: I'll apply this patch during next rebase/merge.
```suggestion
    from kobo.hub.models import Arch, Worker  # noqa: E402
```

osh/worker/tasks/task_get_analyzer_versions.py
```diff
@@ -17,7 +17,7 @@ class AnalyzerVersionRetriever(TaskBase):
     enabled = True
 
     # list of supported architectures
-    arches = ["noarch"]
+    arches = ["noarch", "aarch64", "x86_64"]
```
kdudka, line 20: Does it make sense to maintain a separate list of arches per each task class?  Could we rather move the list of supported arches to `worker.conf` so that OSH admins can introduce new arches without changing the code of OSH?

kdudka (approved):

osh/worker/tasks/task_get_analyzer_versions.py
lzaoral: That depends, now the settings can be more granular.  Nevertheless, such feature would require a modification in `kobo`.

On a second thought, such setting may be confusing.  It would not say what arches are supported by the `worker` itself, that is determined by the `hub`.  It only says that that the implementation of that task supports given arches in general, i.e. it has to be modified only if a new arch is introduces/removed to the whole service deployment.

scripts/add-worker-arch.py
lzaoral: Addressed in 7b998de.

osh/worker/tasks/task_get_analyzer_versions.py
lzaoral: I still may change the `BaseTaskSettings` class name, though.

kdudka: What about naming it `AbstractTask` and letting it inherit from `TaskBase` so that we do not need to use multiple inheritance on all the leaf classes?

lzaoral: Addressed in c6676c9.  Setting `enabled` is necessary so that worker does not see intermediate classes as available task types (the example uses a patched worker printing all registered types):
```sh
$ podman start -a osh-worker
ShutdownWorker TaskBase DiffBuild MockBuild VersionDiffBuild AnalyzerVersionRetriever ErrataDiffBuild
...
```

osh/worker/tasks/task_build.py
```diff
@@ -93,19 +109,13 @@ def notification(cls, hub, conf, task_info):
...
 
-class DiffBuild(Build, TaskBase):
-    def __init__(self, *args, **kwargs):
-        Build.__init__(self)
-        TaskBase.__init__(self, *args, **kwargs)
+class DiffBuild(Build):
+    enabled = True
 
 
 class MockBuild(Build, TaskBase):
```
kdudka, line 116: Now we can remove `, TaskBase` also here.

osh/worker/tasks/task_build.py
```diff
@@ -93,19 +109,13 @@ def notification(cls, hub, conf, task_info):
...
-    def __init__(self, *args, **kwargs):
-        Build.__init__(self)
-        TaskBase.__init__(self, *args, **kwargs)
+    enabled = True
 
 
 class VersionDiffBuild(Build, TaskBase):
```
kdudka, line 120: ... and here.

kdudka (commented):
Thanks for the update!  Two minor remarks inline.

osh/worker/tasks/task_build.py
lzaoral: Thank you! I forgot to change that line as well.  Addressed in 9ea6b52/

osh/worker/tasks/task_build.py
lzaoral: Thank you! I forgot to change that line as well.  Addressed in 9ea6b52/

PR #205, opened 2023-12-08
packaging: add a virtual provide for `osh-cli`

kdudka:
Although the package is named `osh-client`, users know it as `osh-cli` because of how the executable is named.  This makes it easier for them to install it.


siteshwar:
It might have been a mistake to use two different names, that is `osh-client` and `osh-cli` for the client.  We should have only used the name `osh-client` for the client part.

siteshwar (approved):

lzaoral (approved):

PR #206, opened 2023-12-11
client: fix zsh path completions

lzaoral:
Complete whole paths if no option is being specified.  Previously, only the first path component would be suggested which broke completion in subdirectories.


osh/client/completion/_osh-cli
```diff
@@ -139,7 +139,7 @@ _osh-cli_args()
         diff-build|mock-build)
             args+=(
                 '(-d --download-results)'{-d+,--download-results=}'[directory for storing results (leave blank for working directory)]:filename:_files -/'
-                '--install=[install packages into chroot wehan scanning tarballs]:filename:_files'
+                '--install=[install packages into chroot whan scanning tarballs]:filename:_files'
```
kdudka, line 142: Do you intentionally spell it `whan` instead of `when`?

kdudka (approved):
I confirm this fixes a pretty annoying bug.

osh/client/completion/_osh-cli
lzaoral: Not at all.  Thank you for spotting that!

lzaoral: Addressed in 1251014e80be949285ff7ba66ff9cd96418620ae.

PR #207, opened 2023-12-12
ci: enable SELinux in Testing Farm jobs

lzaoral:


lzaoral:
@kdudka The CI does not test any file uploading features (yet), so that's why it is green even with SELinux enabled.

kdudka (approved):
Looks good.  Thanks!

PR #220, opened 2024-01-03
containers+hub: minor improvements

lzaoral:
* containers: use containers/hub/run.sh from the host
* containers: execute hub/run.sh with set -ex
* hub: make display of search boxes consistent

### DIFF
--- a/osh/hub/templates/waiving/list.html
+++ b/osh/hub/templates/waiving/list.html
@@ -8,7 +8,9 @@
 {{ search_form.release.label }} {{ search_form.release }}
 {{ search_form.search }}
 <input type="submit" value="{% trans "Search" %}" />
-{% if user.is_authenticated %}<br />{{ search_form.my }} Show only my scans{% endif %}
+{% if user.is_authenticated %}
+{{ search_form.my }} Show only my scans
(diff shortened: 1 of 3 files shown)


osh/hub/templates/waiving/list.html
```diff
@@ -8,7 +8,9 @@ <h2>{% trans title %}</h2>
 {{ search_form.release.label }} {{ search_form.release }}
 {{ search_form.search }}
 <input type="submit" value="{% trans "Search" %}" />
-{% if user.is_authenticated %}<br />{{ search_form.my }} Show only my scans{% endif %}
+{% if user.is_authenticated %}
```
kdudka, line 11: Should we rather move the `<br/>` tag before the condition?  The `Show only my tasks` checkbox at `/task/` is also shown on a separate line.

kdudka (approved):

osh/hub/templates/waiving/list.html
lzaoral: But `/scan/packages/ shows everything on the same line... I guess it does not really matter as long as the single view itself stays consistent.

PR #226, opened 2024-01-15
worker: `osh-worker.service` improvements

lzaoral:
Highlights:
* worker: add support for service reloading
* worker: set service `KillMode` to `mixed`

### DIFF
--- a/osh.spec
+++ b/osh.spec
@@ -68,7 +68,7 @@ Requires: file
 Requires: koji
 Requires: python3-kobo-client
 Requires: python3-kobo-rpmlib
-Requires: python3-kobo-worker >= 0.32.0
+Requires: python3-kobo-worker >= 0.36.1
 Requires: %{name}-common = %{version}-%{release}
 Recommends: osh-worker-conf
 
--- a/osh.spec
+++ b/osh.spec
@@ -224,8 +224,15 @@ fi
 %systemd_preun osh-worker.service
 
 %postun worker
-# osh-worker does not implement reload and restart would interrupt running tasks
-%systemd_postun osh-worker.service
+%if 0%{?fedora} || 0%{?rhel} > 9
+%systemd_postun_with_reload osh-worker.service
+%else
+# Reload service on package upgrade.
+if [ $1 -ge 1 ]; then
+    # Service reloads using systemd-update-helper are broken on RHEL 9.
+    systemctl reload osh-worker.service || :
+fi
+%endif
 
 %files worker-conf-devel
 %attr(640,root,root) %config(noreplace) %{_sysconfdir}/osh/worker.conf
--- a/osh/worker/osh-worker.service
+++ b/osh/worker/osh-worker.service
@@ -5,6 +5,7 @@ After=network-online.target
 
 [Service]
 Type=exec
+ExecReload=kill -HUP $MAINPID
 ExecStartPre=test -r /etc/osh/worker.conf
 ExecStart=/usr/sbin/osh-worker -f
 KillMode=process
--- a/osh/worker/osh-worker.service
+++ b/osh/worker/osh-worker.service
@@ -8,7 +8,7 @@ Type=exec
 ExecReload=kill -HUP $MAINPID
 ExecStartPre=test -r /etc/osh/worker.conf
 ExecStart=/usr/sbin/osh-worker -f
-KillMode=process
+KillMode=mixed
 Restart=on-failure
 StandardOutput=null
 
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ report bugs and feature requests on GitHub using the above URL.
 - `koji`
 - `python3-kobo-client`
 - `python3-kobo-rpmlib`
-- `python3-kobo-worker >= 0.32.0`
+- `python3-kobo-worker >= 0.36.1`
 
 ### client:
 - `koji`


lzaoral:
@kdudka If you want to, you can review these changes already.  They are noop unless https://github.com/release-engineering/kobo/pull/245 is merged.

kdudka:
I tried to send `SIGHUP` to an OSH worker running with stable `kobo` and the process was terminated by the signal.

lzaoral:
My bad, for some reason I thought that the default action for `SIGHUP` is to discard it.

lzaoral:
Kobo 0.36.0 was released and I've built the packages for Fedora and EPEL:  https://bodhi.fedoraproject.org/updates/?search=kobo-0.36.0

Karma is welcome!

kdudka:
The Testing Fram CI jobs are failing with:
```
[...]
+ cp osh/hub/settings_local.ci.py '/usr/lib/python3.*/site-packages/osh/hub/settings_local.py'
cp: cannot create regular file '/usr/lib/python3.*/site-packages/osh/hub/settings_local.py': No such file or directory
```

I think it means that `osh-hub-conf-devel` was not installed for some reason.

osh.spec
```diff
@@ -225,8 +225,7 @@ fi
 %systemd_preun osh-worker.service
 
 %postun worker
-# osh-worker does not implement reload and restart would interrupt running tasks
-%systemd_postun osh-worker.service
+%systemd_postun_with_reload osh-worker.service
```
kdudka, line 228: This does not work as expected on RHEL-9.  On a subsequent update of the `osh-worker` package I can see:
```
  Running scriptlet: osh-worker-0.9.7.20240322.150052.ga358acc7.pr_226-1.el9.noarch                                                                                                      4/6 
/var/tmp/rpm-tmp.uo5laJ: line 1: fg: no job control
warning: %postun(osh-worker-0.9.7.20240322.150052.ga358acc7.pr_226-1.el9.noarch) scriptlet failed, exit status 1

Error in POSTUN scriptlet in rpm package osh-worker
```

When I download the corresponding RPM package, I can see the `%systemd_postun_with_reload` macro unexpanded in the script:
```
% rpm -qp osh-worker-0.9.7.20240322.150052.ga358acc7.pr_226-1.el9.noarch.rpm --scripts
warning: osh-worker-0.9.7.20240322.150052.ga358acc7.pr_226-1.el9.noarch.rpm: Header V4 RSA/SHA256 Signature, key ID 00cf68b9: NOKEY
postinstall scriptlet (using /bin/sh):

 
if [ $1 -eq 1 ] && [ -x "/usr/lib/systemd/systemd-update-helper" ]; then 
    # Initial installation 
    /usr/lib/systemd/systemd-update-helper install-system-units osh-worker.service || : 
fi
preuninstall scriptlet (using /bin/sh):

 
if [ $1 -eq 0 ] && [ -x "/usr/lib/systemd/systemd-update-helper" ]; then 
    # Package removal, not upgrade 
    /usr/lib/systemd/systemd-update-helper remove-system-units osh-worker.service || : 
fi
postuninstall scriptlet (using /bin/sh):
%systemd_postun_with_reload osh-worker.service
^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

kdudka (changes requested):

jamacku (approved):
LGTM

osh.spec
```diff
@@ -225,8 +225,15 @@ fi
...
 %postun worker
-# osh-worker does not implement reload and restart would interrupt running tasks
-%systemd_postun osh-worker.service
+%if 0%{?fedora} || 0%{?rhel} > 9
+%systemd_postun_with_reload osh-worker.service
+%else
+# Needs systemd 248 or newer
+if [ $1 -ge 1 ] && [ -x "/usr/lib/systemd/systemd-update-helper" ]; then
+    # Package upgrade, not uninstall
+    systemctl set-property osh-worker.service Markers=+needs-reload || :
```
kdudka, line 234: It does not work as expected.  I tried it on a staging instance and it **restarted** the worker instead of reloading:
```
Mar 28 15:23:27 osh-test-worker-001.osh-001.preprod.iad2.dc.redhat.com systemd[1]: Queuing reload/restart jobs for marked units…
Mar 28 15:23:27 osh-test-worker-001.osh-001.preprod.iad2.dc.redhat.com systemd[1]: Stopping OpenScanHub worker daemon...
Mar 28 15:23:28 osh-test-worker-001.osh-001.preprod.iad2.dc.redhat.com systemd[1]: osh-worker.service: Deactivated successfully.
Mar 28 15:23:28 osh-test-worker-001.osh-001.preprod.iad2.dc.redhat.com systemd[1]: Stopped OpenScanHub worker daemon.
Mar 28 15:23:28 osh-test-worker-001.osh-001.preprod.iad2.dc.redhat.com systemd[1]: osh-worker.service: Consumed 1h 5min 39.936s CPU time.
Mar 28 15:23:28 osh-test-worker-001.osh-001.preprod.iad2.dc.redhat.com systemd[1]: Starting OpenScanHub worker daemon...
Mar 28 15:23:28 osh-test-worker-001.osh-001.preprod.iad2.dc.redhat.com systemd[1]: Started OpenScanHub worker daemon.
```

lzaoral: Hmm...  Either `+needs-reload` marker is not processed correctly on RHEL 9, or `systemd` has not reloaded its own configuration using `daemon-reload` yet so it does not know yet about the added `ExecReload:` line in the unit file.  Internally, the `/usr/lib/systemd/systemd-update-helper` script calls `systemctl try-reload-or-restart` after the update is finished:
```
try-reload-or restart PATTERN...
           Reload one or more units if they support it. If not, stop and then start them instead. This does nothing if the units are not running.
```
so that explains the restart but I would expect this script to call `daemon-reload` first before processing marked units.

@kdudka Could you please try calling `systemctl daemon-reload` and then reinstalling the current `osh-worker` package to see it gets then reloaded correctly?

kdudka: I think that @jamacku looked already and submitted https://github.com/redhat-plumbers/systemd-rhel9/pull/250 based on this.  But to me it seems like an overkill.  Cannot we just invoke `systemctl reload osh-worker.service` in the RPM scriptlet on RHEL-9?

jamacku: @lzaoral, I have just noticed that RHEL8 is running `v239`

kdudka: > systemctl reload should be called after systemctl daemon-reload to use the latest unit file.

I would not worry about that.  The unit file is changed approx. twice a year and the `reload` operation would be useful even if it ran on the outdated version of the unit file.

lzaoral: Addressed in e210c3f7581836129341f5a37f6561a6f23a2c8a.

osh/worker/osh-worker.service
```diff
@@ -5,9 +5,10 @@ After=network-online.target
 
 [Service]
 Type=exec
+ExecReload=kill -HUP $MAINPID
 ExecStartPre=test -r /etc/osh/worker.conf
 ExecStart=/usr/sbin/osh-worker -f
-KillMode=process
+KillMode=mixed
```
kdudka, line 11: I find this part of the commit message of c98d498bb8e7ae47481f38d4387a8e1ec761d126 misleading:
```
With the current value of `KillMode=`, ...
```

I guess `current value` refers to the value before the change.  So I would either say `previous value` or expand the value directly:
```
With `KillMode=process` ...
```

kdudka (approved):
It works as expected now.  Thanks!

osh/worker/osh-worker.service
lzaoral: Addressed in 6731c38.

PR #228, opened 2024-01-15
hub: implement dynamic generation of mock configs

lzaoral:
~~Depends on: #221~~

### DIFF
--- a/osh/hub/scan/fixtures/initial_data.json
+++ b/osh/hub/scan/fixtures/initial_data.json
@@ -271,6 +271,14 @@
   }
 },
 {
+  "pk": 7,
+  "model": "scan.mockconfig",
+  "fields": {
+    "enabled": true,
+    "name": "auto"
+  }
+},
+{
   "pk": 9,
   "model": "scan.mockconfig",
   "fields": {
--- a/osh/hub/scan/mock.py
+++ b/osh/hub/scan/mock.py
@@ -1,0 +1,159 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+# SPDX-FileCopyrightText: Copyright contributors to the OpenScanHub project.
+
+"""
+Functions related to dynamic generation of mock configs
+"""
+
+import logging
+import os
+import re
+import subprocess
+import tempfile
+import uuid
+
+import koji
+from django.conf import settings
+
+logger = logging.getLogger(__name__)
+
+
+def _get_available_parent_target(koji_proxy, target):
+    """
+    returns the first usable parent build target, otherwise returns None
+    """
+    parents = koji_proxy.getFullInheritance(target)
+    while parents:
+        target = parents[0]['name']
+        if koji_proxy.getBuildTarget(target):
+            return target
+        parents = koji_proxy.getFullInheritance(target)
+
+    return None
+
+
+def _get_build_method_build_tag(koji_proxy, nvr, target):
+    """
+    returns build tag name of the first usable parent build target for given
+    task, otherwise returns None
+    """
+    # rawhide buildroot always points to latest future release which may be
+    # too new for old builds
+    if target == 'rawhide':
+        tags = sorted(t['name'] for t in koji_proxy.listTags(nvr, pattern='f*'))
+        # use lowest 'f*' tag
+        target = tags[0] if tags else None
+
+    # handle the case if the target repo is no longer available
+    if target is not None and koji_proxy.getBuildTarget(target) is None:
+        # try to use the closest available parent (e.g. for merged side tags)
+        target = _get_available_parent_target(koji_proxy, target)
+
+    if target is None:
+        return
+
+    # get build tag name (build tag seems to always contain the release version)
+    return koji_proxy.getBuildTarget(target)['build_tag_name']
+
+
+def _get_build_arches(koji_proxy, task_id):
+    """
+    returns all arches available for given build
+    """
+    arches = []
+    for child_id in koji_proxy.getTaskDescendents(task_id, request=True):
+        child = koji_proxy.getTaskInfo(child_id, request=True)
+        if child['method'] == 'buildArch' and child['arch'] != 'noarch':
+            arches.append(child['arch'])
+
+    # noarch tasks can be built anywhere
+    return arches or koji_proxy.getAllArches()
+
+
+def _get_tag_specific_config(target):
+    """
+    returns build tag specific mock configuration which should be appended to
+    the generated mock config
+    """
+    # CentOS Stream
+    match = re.search(r'c(\d+)s', target)
+    if match is not None:
+        return f"config_opts['releasever'] = '{match[1]}'\n" \
+               f"config_opts['bootstrap_image'] = 'quay.io/centos/centos:stream{match[1]}'\n"
+
+    # Fedora
+    match = re.search(r'f(\d+)', target)
+    if match is not None:
+        return f"config_opts['releasever'] = '{match[1]}'\n" \
+               f"config_opts['bootstrap_image'] = 'registry.fedoraproject.org/fedora:{match[1]}'\n"
+
+    # RHEL
+    match = re.search(r'rhel-?(\d+)', target)
+    if match is not None:
+        return f"config_opts['releasever'] = '{match[1]}'\n" \
+               f"config_opts['bootstrap_image'] = 'registry.access.redhat.com/ubi{match[1]}/ubi'\n"
+
+    # fallback
+    return "config_opts['use_bootstrap_image'] = False\n"
+
+
+def _create_mock_config(tag, arch, koji_profile, dest_dir):
+    """
+    creates a single mock config for given tag and architecture and stores it
+    in given directory
+    """
+    p = subprocess.run(['koji', '-p', koji_profile, 'mock-config', '--latest',
+                        '--arch', arch, '--tag', tag],
+                       check=True, stdout=subprocess.PIPE)
+    contents = p.stdout.decode()
+
+    # FIXME: remove when dnf5 is usable with mock
+    if 'dnf5' in contents:
+        contents = contents.replace('dnf5', 'dnf')
+
+    # add extra repos
+    matched_repos = [repo for regex, repo
+                     in getattr(settings, 'MOCK_AUTO_EXTRA_REPOS', {}).items()
+                     if re.search(regex, tag)]
+    extra_repos = '\n'.join(matched_repos).replace('\n', '\\n')
+    contents = contents.replace("[main]", extra_repos + '\\n\\n[main]')
+
+    # use randomly generated root directory name
+    contents = re.sub(r"(config_opts\['root'\] = '[^']*)'",
+                      fr"\1-{uuid.uuid4()}'", contents)
+
+    with open(os.path.join(dest_dir, f'mock-{arch}.cfg'), 'w') as f:
+        f.write(contents)
+        f.write(_get_tag_specific_config(tag))
+
+
+def generate_mock_configs(nvr, koji_profile):
+    """
+    returns a path to temporary directory with generated mock configs
+    for given NVR
+    """
+    tmpdir = tempfile.mkdtemp()
+
+    # retrieve the build
+    cfg = koji.read_config(koji_profile)
+    koji_proxy = koji.ClientSession(cfg['server'])
+    build = koji_proxy.getBuild(nvr)
+
+    # retrieve task parameters
+    task = koji_proxy.getTaskInfo(build['task_id'], request=True)
+    method = task['method']
+    params = koji.parse_task_params(method, task['request'])
+
+    # parse build tag name from task parameters
+    if method == 'build':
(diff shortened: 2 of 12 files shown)


lzaoral:
TODO: Add a fixture for the `auto` mock config + Add some tests.

lzaoral:
There is something completely wrong with the coverage gathered by the integration tests.  A file that was definitely executed, is shown as not covered: https://app.codecov.io/gh/openscanhub/openscanhub/commit/02c9f4265ea2a0a471d1abe2ddb521655e90a63e/blob/osh/hub/scan/mock.py

lzaoral:
This still needs some extra work...
* Fail consistently for tasks scheduled by ET, Web-UI and osh-cli.
* At the moment, it is not possible to fail tasks in the `CREATED` state.

lzaoral:
After discussion with @kdudka, I've decided to not implement the features in https://github.com/openscanhub/openscanhub/pull/228#issuecomment-1908162416 due to their complexity and to do that later.  Besides, the `auto` mock config is a very experimental feature anyway.

kdudka:
@lzaoral Thansk!  I tried it out and it worked as expected.

Do I understand it correctly that the API-breaking change from #221 was reworked and the changes for hub/worker can now be deployed per partes?

osh/hub/scan/fixtures/initial_data.json
```diff
@@ -270,6 +270,14 @@
     "name": "fedora-40-x86_64"
   }
 },
+{
+  "pk": 7,
+  "model": "scan.mockconfig",
+  "fields": {
+    "enabled": true,
```
lzaoral, line 277: On a second thought, I'll disable this feature by default because it is quite experimental.
```suggestion
    "enabled": false,
```

kdudka: I would not worry too much about it because OSH instances with a fresh database are currently used only for experiments anyway.  If you disable it in the fixture, you might need to re-enable it in the CI scripts to keep the CI working.

lzaoral: Alright then! 😄

osh/hub/scan/mock.py
```diff
@@ -0,0 +1,133 @@
...
+    logger.debug(f'Generating mock configs for build tag "{tag}"')
+    for arch in _get_build_arches(koji_proxy, task['id']):
+        p = subprocess.run(['koji', '-p', koji_profile, 'mock-config',
+                            '--arch', arch, '--latest', '--tag', tag],
+                           check=True, stdout=subprocess.PIPE)
+        contents = p.stdout.decode()
+
+        # add extra repos
+        matched_repos = [repo for regex, repo
+                         in settings.MOCK_AUTO_EXTRA_REPOS.items()
```
kdudka, line 122: Could we please make it default to an empty dict so that it does not traceback in case the code is updated but local changes in `settings_local.py` are not yet merged?
```py
% osh-cli mock-build --config=auto --nvr curl-7.87.0-7.fc38
  File "/usr/bin/osh-cli", line 85, in <module>
    main()
  File "/usr/bin/osh-cli", line 78, in main
    parser.run()
  File "/usr/lib/python3.11/site-packages/kobo/cli.py", line 296, in run
    cmd.run(*cmd_args, **cmd_kwargs)
  File "/usr/lib/python3.11/site-packages/osh/client/commands/cmd_build.py", line 175, in run
    handle_perm_denied(e, self.parser)
  File "/usr/lib/python3.11/site-packages/osh/client/commands/cmd_build.py", line 173, in run
    task_id = self.submit_task(options)
              ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/osh/client/commands/cmd_mock_build.py", line 20, in submit_task
    return self.hub.scan.mock_build(options['mock_config'],
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/xmlrpc/client.py", line 1122, in __call__
    return self.__send(self.__name, args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/xmlrpc/client.py", line 1464, in __request
    response = self.__transport.request(
               ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/kobo/xmlrpc.py", line 477, in request
    result = transport_class.request(self, *args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/xmlrpc/client.py", line 1166, in request
    return self.single_request(host, handler, request_body, verbose)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/kobo/xmlrpc.py", line 391, in _single_request3
    return self.parse_response(response)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/xmlrpc/client.py", line 1354, in parse_response
    return u.close()
           ^^^^^^^^^
  File "/usr/lib64/python3.11/xmlrpc/client.py", line 668, in close
    raise Fault(**self._stack[0])
xmlrpc.client.Fault: <Fault 1: AttributeError: 'Settings' object has no attribute 'MOCK_AUTO_EXTRA_REPOS'>
```

osh/hub/scan/mock.py
```diff
@@ -0,0 +1,133 @@
...
+                         if re.search(regex, tag)]
+        extra_repos = '\n'.join(matched_repos).replace('\n', '\\n')
+        contents = contents.replace("[main]", extra_repos + '\\n\\n[main]')
```
kdudka, line 125: Could we please make `config_opts['root']` unique as well?  The locking mechanism in `csmock` assumes that each mock config uses its own directory for the (ch)root.  If we run two distinct mock configs that use the same value for `config_opts['root']`, the locking mechanism does not work:
```
[...]
>>> 2024-02-19 14:29:12	"/usr/bin/mock" "-r" "/tmp/tmpyfvruenl/mock.cfg" "--plugin-option=tmpfs:keep_mounted=True" "--config-opts=print_main_output=True" "--init"
INFO: mock.py version 5.2 starting (python version = 3.9.18, NVR = mock-5.2-1.kdudka.5.el9), args: /usr/libexec/mock/mock -r /tmp/tmpyfvruenl/mock.cfg --plugin-option=tmpfs:keep_mounted=True --config-opts=print_main_output=True --init
Start(bootstrap): init plugins
INFO: selinux enabled
Finish(bootstrap): init plugins
Start: init plugins
INFO: selinux enabled
Finish: init plugins
INFO: Signal handler active
Start: run
Start: clean chroot
ERROR: Build root is locked by another process.

!!! 2024-02-19 14:29:13	warning: failed to init mock profile (/tmp/tmpyfvruenl/mock.cfg), trying to scrub cache...
[...]
```

Rather than fixing the locking mechanism in `csmock` and making the tasks block each other, we can use separate (ch)roots and make the tasks finally run in parallel.

lzaoral: In that case, we must implement some clean-up logic, because some chroot remains will be left on the filesystem when the task finishes.

osh/hub/scan/mock.py
lzaoral: And what about doing it the other way around?  First contribute you downstream settings to `osh-rh-conf` and then do the update.

kdudka: Technically, it can be done.  My impression is that handling of configuration files in other projects that I use does not require each supported option to be present in each configuration file.  When an option has a reasonable default value, the application should simply use the default value when the option is missing in a config file.

lzaoral: Ok, I'll move the default to `settings.py`.

osh/hub/scan/mock.py
kdudka: Good point.  `csmock` should optionally invoke `mock ... --scrub=all` before releasing the lock.  Something like https://github.com/openscanhub/openscanhub/blob/77f43ad4d0c0860ce1b72f9edae6717847a66003/osh/worker/csmock_runner.py#L123

In any case, this problem will go away once OSH starts to create a fresh worker for each task (similarly to how COPR works).

osh/hub/scan/mock.py
```diff
@@ -0,0 +1,133 @@
...
+        if koji_proxy.getBuildTarget(target):
+            return target
+        parents = koji_proxy.getFullInheritance(target)
+
+    return None
+
+
+def _get_build_method_build_tag(koji_proxy, target):
+    """
+    returns build tag name of the irst usable parent build target for given
```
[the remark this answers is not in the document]
lzaoral: Addressed in ea34cbc3cef206f26c4d5fad92232cbc86a02e98.

osh/hub/scan/mock.py
lzaoral: Random root directory names are handled in ea34cbc3cef206f26c4d5fad92232cbc86a02e98.

lzaoral: Both `csmock` pull requests are now merged.  I'll add the support for `--scrub-all` ASAP.

lzaoral: Addressed in 74e85ca87ac515862ac2996fec87ee353acbddb4.

lzaoral: The jobs will fail because there has not been a release of `csmock` yet.

osh/hub/scan/mock.py
kdudka: @lzaoral Please put there at least the fallback to an empty dict.  Even if we update `osh-rh-conf`, the changes need to be manually merged because of [OSH-387](https://issues.redhat.com/browse/OSH-387).

kdudka (approved):
Works as expected.  Thanks!

lzaoral:
With c38e870, all TF tests pass so there must be some SELinux-related issue.

osh/hub/scan/mock.py
lzaoral: Addressed in 75eb823.

osh/hub/scan/scanner.py
```diff
@@ -38,6 +39,18 @@ def dig_arch(mock_config):
     return 'noarch'
 
 
+# exist but raises SELinux if it craetes the target directory on its own
```
rhyw, line 42: typo here

osh/worker/tasks/task_build.py
```diff
@@ -54,20 +55,28 @@ def run(self):
             self.spawn_subtask(*base_task_args, inherit_worker=True)
             self.wait()
 
+        # download custom mock config from the hub
+        mock_config_url = None
+        if mock_config == 'auto':
```
rhyw, line 60: do you think it necessary to update the command line helper to make it explicit? `mock-build`, `diff-build` etc. If I understand correctly it still requires to specify `--config=auto` from client otherwise falls to the default mock config which is populated from `DEFAULT_MOCKCONFIG` from client or system conf settings.

Moreover should we promote `DEFAULT_MOCKCONFIG = auto` from now on? Just out of curiosity.

lzaoral: I'd rather keep this feature hidden for now because it is still quite experimental.

> Moreover should we promote DEFAULT_MOCKCONFIG = auto from now on? Just out of curiosity.

I don't think so, this feature needs actual Koji/Brew builds to exist so it is incompatible with SRPMs uploaded through the client.

kdudka: I believe that `--config=auto` will be the default one day.  But a few more steps need to be completed to get there.  We need to introduce some automagic rules to guess the correct profile for uploaded SRPMs and to use `cspodman` for Brew builds of containers (and ideally move the cspodman-specific bits to `osh-rh-conf`).  For now we need to stabilize the initial implementation before we proceed with the next steps.

osh/worker/csmock_runner.py
```diff
@@ -123,6 +119,10 @@ def analyze(self, analyzers, srpm_path, profile=None, su_user=None, additional_a
             cmd += " --force-global-cleanup-on-exit"
         else:
             cmd = "csmock"
+            if profile_url:
+                profile = self.download_file(profile_url, 'mock.cfg')
+                # do not cache anything when custom mock profiles are used
+                cmd += ' --scrub-on-exit '
```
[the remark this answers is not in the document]
lzaoral: Addressed in 07f6cc16d97fa4fbfc41033982d395be2d1ef1a0.

osh/hub/scan/scanner.py
lzaoral: Thank you for spotting it!  Somehow, I have committed an incomplete comment. Fixed in 361bf3c5ba485fa2afa81fdba44966c833caad5e.

rhyw (approved):

kdudka (approved):

PR #229, opened 2024-01-23
ci: improve coverage by restarting/stopping containers properly

lzaoral:
Related: https://github.com/openscanhub/openscanhub/pull/228#issuecomment-1906123346


lzaoral:
As it turns out, the code that is execute by `manage.py runserver` in the integration test suite is not measured at all at the moment.

@siteshwar You may look into this in the future ^.

PR #234, opened 2024-02-21
scripts: add script to spin up single use workers

siteshwar:


osh/hub/settings_local.py
```diff
@@ -130,4 +130,9 @@
 # E-mail address where notifications about failed/cancelled tasks are sent
 DEVEL_EMAIL_ADDRESS = "no-reply@example.org"
 
+# TODO: What should we use here if we want to deploy across multiple clouds?
+RESALLOC_WORKER_SSH_PRIVATE_KEY = "id_rsa"
+OSH_HUB_URL = "http://ec2-54-202-223-253.us-west-2.compute.amazonaws.com:8000/xmlrpc"
+ENABLE_THROWAWAY_WORKERS = True
```
siteshwar, line 136: Shall we name this setting differently? `OSH_HUB_URL` and `RESALLOC_WORKER_SSH_PRIVATE_KEY` are only used if this setting is enabled.

osh/hub/osh_xmlrpc/worker.py
```diff
@@ -11,6 +11,11 @@
 from kobo.hub.decorators import validate_worker
 from kobo.hub.models import Task
 
+from kobo.hub.xmlrpc.worker import open_task as kobo_open_task
+
+from osh.hub.errata.models import ScanningSession
```
lzaoral, line 16: This model was removed in https://github.com/openscanhub/openscanhub/pull/170 at the and of November 2023.  Please, make sure your PR is up-to-date with `main`.

lzaoral (changes requested):

osh/hub/scripts/osh-resalloc
```diff
@@ -0,0 +1,133 @@
...
+
+def delete_worker(ip_address):
+    from kobo.hub import models
+    # Worker name should be set to hostname of the worker
+    worker_name = get_hostname_from_ip(ip_address)
+    kobo_worker = models.Worker.objects.get(name=worker_name)
+    kobo_worker.delete()
+
+def workers_needed():
+    print(8)
```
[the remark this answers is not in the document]
kdudka: You mean the number of pre-allocated workers?  That could be a configuration option I guess.

osh/hub/settings_local.py
kdudka: Do we need to manually configure `OSH_HUB_URL`?  Cannot the information be obtained from Django and/or Apache?  Could we at least use some automagically obtained value as a default to cover  the most common cases?

What about renaming `ENABLE_THROWAWAY_WORKERS` to `ENABLE_SINGLE_USE_WORKERS`?

osh/hub/scripts/osh-resalloc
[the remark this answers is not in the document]
siteshwar: Yes, but the number of workers should be determined dynamically. For example,  `log(number_of_pending_tasks)` as long as it's below a certain maximum limit. We should only run certain number of workers (like 2) as pre allocated workers.

osh/hub/settings_local.py
siteshwar: > Do we need to manually configure OSH_HUB_URL? Cannot the information be obtained from Django and/or Apache? 

> Could we at least use some automagically obtained value as a default to cover the most common cases?

The hostname should be reachable to the worker and is only used when resalloc is enabled. I am not aware what value may be suitable for such case. We can probably rename this setting to `RESALLOC_OSH_HUB_URL` and leave a blank value as default.

kdudka: Yes, we can certainly do that now and eventually figure out a way how to query/guess the correct value when `RESALLOC_OSH_HUB_URL` is not specified in the config file.

osh.spec
```diff
@@ -233,6 +233,7 @@ fi
 
 %files hub
 %defattr(-,root,apache,-)
+%{_sbindir}/osh-resalloc
```
siteshwar, line 236: Is `/sbin` the correct path to store this script?

osh/hub/settings_local.py
siteshwar: On a second note, shall we use the term `RESALLOC` at all in the OpenScanHub source code? We might just use the term `SINGLE_USE_WORKERS` everywhere, just in case we switch to some other backend for running virtual machines dynamically.

osh.spec
kdudka: `%{_sbindir}` expands to `/usr/sbin` on a Fedora system.  Assuming root privileges are needed to run the executable, I think the path is correct.  This Fedora page is closely related: https://fedoraproject.org/wiki/Changes/Unify_bin_and_sbin

siteshwar: This should be run as exactly the same user and group as the webserver. It is mentioned at the top of the script.

osh/hub/settings_local.py
```diff
@@ -130,4 +130,11 @@
 # E-mail address where notifications about failed/cancelled tasks are sent
 DEVEL_EMAIL_ADDRESS = "no-reply@example.org"
 
+# If this setting is enabled, a worker is only used to perform a single task.
+ENABLE_SINGLE_USE_WORKERS = True
```
siteshwar, line 134: This should be set to `False` by default before merging.

osh/hub/scripts/osh-resalloc
```diff
@@ -0,0 +1,159 @@
...
+    # Worker name should be set to hostname of the worker
+    worker_name = get_hostname_from_ip(ip_address)
+    kobo_worker = models.Worker.objects.get(name=worker_name)
+    kobo_worker.delete()
+
+
+def workers_needed():
+    # This number should not exceed `max` field in the `/etc/resallocserver/pools.yaml`
+    # on the resalloc server.
+    MAX_SINGLE_USE_WORKERS = 8
```
siteshwar, line 97: This should be moved to the settings file.

osh/hub/scripts/osh-resalloc
```diff
@@ -0,0 +1,159 @@
...
+
+    # Tasks in opened or assigned state
+    running_tasks = kobo.hub.models.Task.objects.running().count()
+
+    # Tasks waiting to be assigned
+    free_tasks = kobo.hub.models.Task.objects.free().count()
+
+    running_and_free_tasks = running_tasks + free_tasks
+
+    print(max(running_and_free_tasks, MAX_SINGLE_USE_WORKERS))
```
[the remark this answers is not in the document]
kdudka: I think it does but I know nothing about resalloc.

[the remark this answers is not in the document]
siteshwar: It would end up with `MAX_SINGLE_USE_WORKERS` number of workers, even if both running and free tasks are 0. We probably need a better equation here.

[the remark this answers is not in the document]
kdudka: Of course, `max(...)` needs to be replaced with `min(...)`.  My bad.

osh/hub/scripts/osh-resalloc
```diff
@@ -0,0 +1,161 @@
...
+    # and it should be deleted after that.
+    kobo_worker.max_load = 1
+    kobo_worker.arches.add(models.Arch.objects.get(name="x86_64"))
+    kobo_worker.channels.add(models.Channel.objects.get(name="default"))
+    kobo_worker.save()
+    logger.debug(f"Created kobo worker for hostname {worker_name}")
+    return kobo_worker
+
+
+def start_worker(kobo_worker):
```
siteshwar, line 39: Alternatively, content of this function could be put into an ansible playbook which should be run here. This way we would avoid directly running shell commands.

osh.spec
siteshwar: Shall this be moved into a separate subpackage?

kdudka: As long as it does nothing in the default configuration and does not introduce any new RPM-level dependencies, I think it is safe to keep it in the existing subpackage.  @lzaoral What is your opinion on this?

osh/hub/scripts/osh-resalloc
siteshwar: Resalloc has 3 different values to determine number of virtual machines being spun up in [pools.yaml](https://github.com/praiskup/resalloc/blob/337b16d322d11c524d5e4230ed6c91fd77a1f7f4/config/pools.yaml#L5-L22):

- `max` determines number of maximum virtual machines.
- `max_starting` determines number of maximum virtual machines which are starting.
- `max_prealloc` determines number of maximum pre allocated virtual machines.

Shall we find a way to pick up one of those values to determine number of maximum single use workers? If it's not possible we need to leave a comment that `MAX_SINGLE_USE_WORKERS` should never be above `max` number of nodes from the `pools.yaml`, otherwise the agent spawner daemon would be stuck waiting for tickets that may never be resolved and would not clean up workers that have been already used.

osh.spec
siteshwar: If we have to find a way to automatically determine number of running nodes, we may end up creating a dependency on the `resalloc` package. In that case, I would move the script to a separate subpackage.

osh/hub/settings_local.py
siteshwar: We should probably rename this script to `osh-worker-manager` (or something similar). And create a subpackage with same name, if needed.

osh/hub/scripts/osh-resalloc
siteshwar: > That sounds wrong. Inability to allocate more machines (for whatever reason) should not block releasing of machines that become unused. Otherwise the solution is prone to deadlock.

@praiskup I remember that was the limitation of the first prototype that we did. Does that limitation still exist? The code may have changed over time.

praiskup: I am sure I do not fully understand the question, so just "trying" to answer :-)

- agent spawner just tries to converge to the ideal number of agents, IOW it opens as many resalloc tickets as necessary (plus runs scripts against them, once allocated, if configured to do so)
- the fact that the ticket is open doesn't block the spawner logic, neither resalloc, nor osh
- resalloc's pools.yaml `max` value should be >= the `converge` number, just to be able to converge at some point ... but nothing serious happens even if not (the resalloc tickets stay open..., no more resources are allocated, and agent spawner doesn't receive the agents)

Said that, no blocking (anymore?). All the tasks done by agent spawner are async.

I'm not sure why we need `MAX_SINGLE_USE_WORKERS` here, not sure whether we have to "divide" or categorize OSH workers like that at all. The resalloc workers (agents) may be reused, same as any other workers.  Whether they are/are not is a matter of configuring the agent's hooks appropriately (so ideally, OSH doesn't care at all).

osh/hub/scripts/osh-resalloc
```diff
@@ -0,0 +1,162 @@
...
+
+    # Tasks in opened or assigned state
+    running_tasks = kobo.hub.models.Task.objects.running().count()
+
+    # Tasks waiting to be assigned
+    free_tasks = kobo.hub.models.Task.objects.free().count()
+
+    running_and_free_tasks = running_tasks + free_tasks
+
+    print(min(running_and_free_tasks, MAX_SINGLE_USE_WORKERS))
```
praiskup, line 112: Note that you need to wait some nontrivial time till the agent becomes available. So if we combine single-task workers and multi-task workers, some of them might get allocated and never used eventually.

My original thinking was that we could have something like a logarithmic function of free tasks in the queue (or something like that).  I.e. something like `ceil(ln(free_tasks)) + running_tasks`.

praiskup: Perhaps just `ceil(ln(free_tasks))` and do not return `running_tasks`.  If there are no more free tasks, there's no need to preallocate more agents when the old agents are terminated.

kdudka: I understand that allocation of machines takes time but how is it related to the number of **needed** workers?

osh/hub/scripts/osh-resalloc
siteshwar: > the fact that the ticket is open doesn't block the spawner logic, neither resalloc, nor osh

It was initially blocking in the first prototype we did few months ago. I am happy that it has changed.

> I'm not sure why we need `MAX_SINGLE_USE_WORKERS` here, not sure whether we have to "divide" or categorize OSH workers like that at all. The resalloc workers (agents) may be reused, same as any other workers. Whether they are/are not is a matter of configuring the agent's hooks appropriately (so ideally, OSH doesn't care at all).

It may just be a method to limit number of open tickets in resalloc. For example, what happens if a mass scan is initiated for new version of Fedora. However, it would be good to keep this number in sync with `max` value in `pools.yaml`.

osh/hub/scripts/osh-resalloc
praiskup: Extreme example:
```
initial state: two static reusable workers, 0 dynamic (agent spawner), two tasks to be processed
1. the two tasks are taken by static workers, we report "ideal number = 2"
2. agent spawner allocates the two workers
3. tasks are processed, "ideal number = 2"
4. agent spawner downsizes... and terminates agents
```

But even without static workers, if we make the dynamic workers reusable:
```
initial state: 1 dynamic worker allocated, one running task, two tasks in queue
1. ideal number reported = 3
2. spawner starts allocating 2 more agents
3. the running task is finished, the worker is free -> and takes another task
4. the two agents got allocated && ready, there's only one task to finish
```

> I understand that allocation of machines takes time but how is it related to the number of needed workers?

Since the spawner only poorly delivers the requested number of agents that OSH asks for, we can estimate or guess in advance whether the workers will be needed. No big deal, the stack of workers will get downsized anyway... but there's unnecessary allocation/deallocation traffic.

If the particular OSH instance uses only dynamic && single-task workers, we don't have to worry. Combining with static workers is probably not considered in the public OSH instance anyway. However, reusing allocated agents seems like a perfectly valid use case; that's what we actually do in Fedora Copr, too. Builders are recycled for multiple tasks as long as it is safe to do so (the task is owned by the same user, building into the same project, etc.).

osh/hub/scripts/osh-resalloc
siteshwar: I created a subpackage  for `osh-worker-manager` anyways.

osh/hub/scripts/osh-resalloc
praiskup: Yes, I agree and I think you are right (originally we did not plan the spawner would exist, I think?).

Now, I think we should document this constraint, somewhere (either 1. or 2. and the details about 2.).  But note that it is rather an artificial constraint (not sure there are technical reasons to implement it - to have two types of workers, and whether OSH has to know about this at all).

kdudka (changes requested):

osh/hub/scripts/osh-worker-manager
```diff
@@ -0,0 +1,180 @@
...
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "osh.hub.settings")
+    django.setup()
+
+    import argparse
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument("--create-worker", type=str, help="Try to create worker with specified hostname.")
+    parser.add_argument("--delete-worker", help="Delete worker with specified hostname (if its ideal)")
+    parser.add_argument("--workers-needed", action="store_true", help="Count number of queued task and return number of workers needed.")
+    parser.add_argument("--check-finished", type=str, help="Check if the worker should be removed.")
```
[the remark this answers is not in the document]
siteshwar: There is no clear documentation, but you can see a sample file here: https://github.com/praiskup/resalloc/blob/151bd591ca2426b8a6b99253d88761c6e4be9f8c/config/agent-spawner/config.yaml

OSH configurations could be seen here: https://github.com/openscanhub/fedora-infra/blob/e433862e13ec16596b1512249cb0ebfc1133805e/containers/configs/resalloc-agent-spawner-config.yaml

osh/hub/scripts/osh-worker-manager
```diff
@@ -0,0 +1,180 @@
...
+    # Is this safe to assume this would always work?
+    temporary_file.flush()
+    # and transfer it to new worker.
+    # Copy worker.conf to centos user's home directory
+    # This requires the hub to have ssh access to the new worker
+    print(f"scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o ConnectTimeout=20 -i {settings.SINGLE_USE_WORKER_SSH_PRIVATE_KEY} {temporary_file.name} centos@{kobo_worker.name}:worker.conf")
+    scp_exit_status = os.system(f"scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o ConnectTimeout=20 -i {settings.SINGLE_USE_WORKER_SSH_PRIVATE_KEY} {temporary_file.name} centos@{kobo_worker.name}:worker.conf")
+    try:
+        if scp_exit_status != 0:
+            print(f"Failed to copy worker.conf to new worker {kobo_worker.name}")
```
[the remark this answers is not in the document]
siteshwar: They are logged in a separate file for each command execution. Example: https://github.com/openscanhub/fedora-infra/blob/e433862e13ec16596b1512249cb0ebfc1133805e/containers/configs/resalloc-agent-spawner-config.yaml#L15

I plan to keep them in `/var/log/resalloc-agent-spawner` directory.

osh/hub/scripts/osh-worker-manager
```diff
@@ -0,0 +1,180 @@
...
+    # For example, through ansible or paramiko.
+    import sys
+
+    from django.conf import settings
+    from django.template.loader import render_to_string
+
+    # Generate /etc/osh/worker.conf through jinja template
+    worker_conf = render_to_string("worker.conf.j2", {"SINGLE_USE_WORKER_OSH_HUB_URL": settings.SINGLE_USE_WORKER_OSH_HUB_URL, "OSH_WORKER_KEY": kobo_worker.worker_key})
+    # Create a temporary file and save the template
+    temporary_file = tempfile.NamedTemporaryFile()
```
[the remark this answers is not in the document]
siteshwar: These are the permissions I see in the container on the temporary file:

```
-rw-------. 1 1001320000 root 1.3K Mar 18 12:43 tmpvddg64kb
```

So  it should be safe.

osh/hub/scripts/osh-worker-manager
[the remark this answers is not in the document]
siteshwar: I have modified the code to use `stderr`. I would leave a todo note to use python3 logger.

osh/hub/scripts/osh-worker-manager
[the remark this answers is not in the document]
kdudka: Thanks!  So I would put here a comment with reference to the corresponding OSH configuration.

lzaoral (changes requested):
Note that I have just skimmed through the code.  I still have to a proper review.

osh/hub/osh_xmlrpc/worker.py
```diff
@@ -243,3 +245,17 @@ def interrupt_tasks(request, task_list):
...
+@validate_worker
+def open_task(request, task_id):
+    response = kobo_open_task(request, task_id)
+
+    if settings.ENABLE_THROWAWAY_WORKERS:
+        task = Task.objects.get(id=task_id)
+        # Set the worker load to 0. This should avoid getting any new tasks assigned the worker.
+        # Each worker allocated through resalloc should be destroyed once a task completes.
+        task.worker.max_load = 0
+        task.worker.save()
```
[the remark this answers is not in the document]
siteshwar: We agreed with @kdudka to avoid mixing static and dynamically allocated workers. Please see [this comment](https://github.com/openscanhub/openscanhub/pull/234#discussion_r1513029712).

osh/hub/scripts/osh-worker-manager
```diff
@@ -0,0 +1,180 @@
...
+    from django.core.exceptions import ObjectDoesNotExist
+    from kobo.hub import models
+
+    # Check if a worker with this name already exists
+    try:
+        kobo_worker = models.Worker.objects.get(name=worker_name)
+    except ObjectDoesNotExist:
+        # If not, create a new one
+        kobo_worker = models.Worker.create_worker(worker_name)
+        kobo_worker.arches.add(models.Arch.objects.get(name="x86_64"))
```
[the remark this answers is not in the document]
siteshwar: https://github.com/openscanhub/openscanhub/blob/55a0101e073122850ab84e0b43af76787c3fc96d/osh/hub/scripts/osh-worker-manager#L39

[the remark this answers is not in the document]
lzaoral: Yes, I see that. But you have no guarantee that given architecture will exist in the database.  Sure, they are in the default fixtures but they are not loaded when you install the packages AFAIK.

Also, the second part of https://github.com/openscanhub/openscanhub/pull/234#discussion_r1519608321 about the `TODO:` was not addressed either.

osh/hub/scripts/osh-worker-manager
```diff
@@ -0,0 +1,203 @@
...
+    if args.delete_worker:
+        print(f"Deleting {args.delete_worker}")
+        ip_address = args.delete_worker
+        delete_worker(ip_address)
+
+    if args.workers_needed:
+        workers_needed()
+
+    if args.check_finished:
+        check_finished(args.check_finished)
```
lzaoral, line 199: For some commands, you use the  redundant `ip_address` variable, for some you do not.  I vote to use the `args.<name>` member directly.

osh/hub/scripts/osh-worker-manager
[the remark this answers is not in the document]
siteshwar: Would a worker that has `noarch` set, pick up a `x86_64` arch task?

lzaoral (changes requested):
Note that I haven't tried to execute the code itself.

osh/hub/scripts/osh-worker-manager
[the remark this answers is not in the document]
lzaoral: If the task has `x86_64` arch set, than no.  The current arch selection works like this: https://github.com/openscanhub/openscanhub/blob/f5e051c4337fed696481bcff11a4c48243cf2313/osh/hub/scan/scanner.py#L34-L38 
Therefore, if the `x86_64` architecture exists in the database, all tasks with `*-x86_64` mock profiles will use it, otherwise they will fall back on `noarch`.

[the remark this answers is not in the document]
lzaoral: @siteshwar FYI,  https://github.com/openscanhub/openscanhub/pull/234#discussion_r1528764481 was written during review, so it is shown before https://github.com/openscanhub/openscanhub/pull/234#discussion_r1528765616 even though my was published alter.

osh/hub/scripts/osh-worker-manager
```diff
@@ -0,0 +1,203 @@
...
+This script was originally written for the resalloc agent spawner daemon.
+But it may be used by any other resource allocation mechanism.
+This should be run as exactly the same user and group as the webserver. For example, `apache`.
+"""
+
+import os
+import socket
+import tempfile
+
+# TODO: We may use python3 logger instead of `print()` statements for debugging.
```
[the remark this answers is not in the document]
siteshwar: I have not tried it.

osh/hub/scripts/osh-worker-manager
```diff
@@ -0,0 +1,203 @@
...
+    worker_name = get_hostname_from_ip(ip_address)
+    kobo_worker = create_worker_in_kobo(worker_name)
+    start_worker(kobo_worker)
+
+
+def delete_worker(ip_address):
+    from kobo.hub import models
+
+    # Worker name should be set to hostname of the worker
+    worker_name = get_hostname_from_ip(ip_address)
```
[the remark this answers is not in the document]
siteshwar: I would prefer to keep it as it is.

osh/hub/scripts/osh-worker-manager
```diff
@@ -0,0 +1,208 @@
...
+    # Generate /etc/osh/worker.conf through jinja template
+    worker_conf = render_to_string("worker.conf.j2",
+                                   {"SINGLE_USE_WORKER_OSH_HUB_URL": settings.SINGLE_USE_WORKER_OSH_HUB_URL,
+                                    "OSH_WORKER_KEY": kobo_worker.worker_key})
+
+    # Create a temporary file and save the template
+    with tempfile.NamedTemporaryFile() as temporary_file:
+        temporary_file.write(worker_conf.strip().encode())
+        temporary_file.flush()
+        # File is closed, but not removed.
```
siteshwar, line 88: This does not work as documented in my testing and I get below error if I close the file here:

```
stat local "/tmp/tmprimblhg2": No such file or directory^M
Failed to copy worker.conf to new worker ec2-54-184-164-118.us-west-2.compute.amazonaws.com
```

siteshwar:
> @siteshwar Could you please stop pushing your commits after every single change you make? It is spamming my mailbox. Thank you! (If you want to resolve each review comment separately, use fixups and push them all at once.)

I would try to reduce them, but I need to verify my changes work in the testing environment in the Openshift deployment.

osh/hub/scripts/osh-worker-manager
kdudka: `temporary_file.flush()` does **not** close the file.  I think it will be closed at the and of the `with ... as` block.  The fact whether the file will be deleted on close is controlled by the `delete_on_close ` argument of `tempfile.NamedTemporaryFile()` and it defaults to `True`.  So I would say it works as documented: https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

kdudka: @siteshwar What exactly does not work as documented?

siteshwar: > @siteshwar What exactly does not work as documented?

I was not setting `delete_on_close=False`. So I missed that in the documentation.

siteshwar: On another note, `delete_on_close` is [python 3.12](https://docs.python.org/3.12/whatsnew/3.12.html#tempfile). Earlier versions only have `delete`.

osh/hub/scripts/osh-worker-manager
```diff
@@ -0,0 +1,220 @@
...
+    try:
+        temporary_file = tempfile.NamedTemporaryFile()
+        temporary_file.write(worker_conf.strip().encode())
+        # TODO: File is closed, but not removed if `delete_on_close` is True.
+        # temporary_file.close()
+        temporary_file.flush()
+        # and transfer it to new worker.
+        # Copy worker.conf to ec2-user user's home directory
+        # This requires the hub to have ssh access to the new worker
+        scp_cmd = f"scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o ConnectTimeout=20 -i {settings.SINGLE_USE_WORKER_SSH_PRIVATE_KEY} {temporary_file.name} ec2-user@{kobo_worker.name}:worker.conf"
```
siteshwar, line 97: The username for ssh should be configurable through the settings. It is not always same for different virtual machines.

osh/hub/osh_xmlrpc/worker.py
```diff
@@ -76,6 +79,20 @@ def finish_task(request, task_id):
...
 
+@validate_worker
+def open_task(request, task_id):
+    response = kobo_open_task(request, task_id)
+
+    if settings.ENABLE_SINGLE_USE_WORKERS:
+        task = Task.objects.get(id=task_id)
+        # Set the worker load to 0. This should avoid getting any new tasks assigned the worker.
+        # Each single use worker should be destroyed once a task completes.
+        task.worker.max_load = 0
```
siteshwar, line 90: This would cause issues with differential scans. The differential scan task would get assigned, but never start as the `max_load` has been set to `0` before the base scan task is started.

siteshwar: In case of differential scans, this should be set to `0` only after "non-base" scan has started.

lzaoral: Did you actually hit this problem or is it just your assumption? I don't think this is a problem.  Otherwise, we would be hitting this all the time on production workers where one task would block others to spawn children.

Waiting tasks are not counted towards the actual load of the worker:
https://github.com/release-engineering/kobo/blob/9af65943b183e2904b47855cbc820b8dda2d5643/kobo/hub/models.py#L212

That said, enjoy the rest of the weekend!

siteshwar: I hit it while doing a [version-diff-build](https://openscanhub.stg.fedoraproject.org/task/33/) scan. It only started after I reset the `max_load` field manually.

lzaoral: Sorry, my previous comment was completely wrong.

osh/hub/scripts/osh-worker-manager
```diff
@@ -0,0 +1,224 @@
...
+    kobo_worker = models.Worker.objects.get(name=worker_name)
+    kobo_worker.delete()
+
+
+def workers_needed():
+    import kobo.hub.models
+    from django.conf import settings
+
+    # Tasks in opened or assigned state
+    running_tasks = kobo.hub.models.Task.objects.running().count()
```
siteshwar, line 157: We should skip subtasks here.

osh/hub/scripts/osh-worker-manager
```diff
@@ -0,0 +1,224 @@
...
+
+    # Tasks waiting to be assigned
+    free_tasks = kobo.hub.models.Task.objects.free().count()
```
siteshwar, line 160: We should skip subtasks here.

osh/hub/osh_xmlrpc/worker.py
```diff
@@ -76,6 +79,26 @@ def finish_task(request, task_id):
...
+        task = Task.objects.get(id=task_id)
+
+        # Do not set `max_load` to 0 in that case, as the base scan would get assigned but never
+        # start as `max_load` has been set to 0.
+        if task.method == "VersionDiffBuild":
+            pass
+        else:
+            # Set the worker load to 0. This should avoid getting any new tasks assigned the worker.
+            # Each single use worker should be destroyed once a task completes.
+            task.worker.max_load = 0
```
siteshwar, line 96: Shall we be setting this field when the task is closed, interrupted or failed? Could there be any race condition there i.e. a worker may pick up a task before `max_load` is set to `0`.

osh/hub/scripts/osh-worker-manager
```diff
@@ -0,0 +1,224 @@
...
+    # Worker name should be set to hostname of the worker
+    worker_name = get_hostname_from_ip(ip_address)
+    kobo_worker = models.Worker.objects.get(name=worker_name)
+
+    # TODO: Check for stale workers that have max_load above 0, but do not have arch
+    # or channel set. They would never be used.
+
+    # TODO: Shall we perform further health checks here? For example, if `osh-worker`
+    # service is running on the worker machine?
+    if kobo_worker.max_load + kobo_worker.current_load > 0:
```
siteshwar, line 179: When doing `VersionDiffBuild`, there is a time difference between when a subtask finishes and main task starts. The worker would get deleted if agent spawner checks this condition and the task would be left in `OPEN` state and never finish.

osh/hub/osh_xmlrpc/worker.py
```diff
@@ -76,6 +79,27 @@ def finish_task(request, task_id):
                 task.fail_task()
 
 
+@validate_worker
+def open_task(request, task_id):
+    response = kobo_open_task(request, task_id)
+
+    if settings.ENABLE_SINGLE_USE_WORKERS:
```
kdudka, line 86: Will this work when `ENABLE_SINGLE_USE_WORKERS` is missing in `settings_local.py`?

kdudka (commented):
I find the use of `stdout`/`stderr` in `worker-manager` inconsistent.  I think it should use `stdout` for output that is processed by the caller and `stderr` for diagnostic/progress output.

osh/hub/osh_xmlrpc/worker.py
siteshwar: No, it would throw `'Settings' object has no attribute` exception.

kdudka: Then it should be fine.  Btw. it seems you forgot to resolve this TODO before clearing the draft flag on this PR: https://github.com/openscanhub/openscanhub/pull/234/files#diff-e1475eca3773eef4cb4c8503a87591c770a7ed2ebf8e08b205c00faae6ab81a2R187

siteshwar: I just added it today and need to add it to https://pagure.io/fedora-infra/ansible/blob/main/f/roles/openshift-apps/openscanhub/templates/settings_local.py

siteshwar: https://pagure.io/fedora-infra/ansible/pull-request/1884

osh/hub/osh_xmlrpc/worker.py
```diff
@@ -76,6 +79,28 @@ def finish_task(request, task_id):
...
+        task = Task.objects.get(id=task_id)
+
+        Task.create_shutdown_task("worker/" + task.worker.name, task.worker.name, kill=False)
```
siteshwar, line 89: @lzaoral This method may worker better, but would spam the tasks view with `ShutdownWorker` tasks. Shall we modify the view to skip such tasks?

osh/hub/osh_xmlrpc/worker.py
```diff
@@ -76,6 +79,28 @@ def finish_task(request, task_id):
...
 
 
+@validate_worker
+def open_task(request, task_id):
+    response = kobo_open_task(request, task_id)
+
+    if settings.ENABLE_SINGLE_USE_WORKERS:
+        task = Task.objects.get(id=task_id)
+
+        Task.create_shutdown_task(task.worker.name, task.worker.name, kill=False)
```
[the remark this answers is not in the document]
lzaoral: ... and setting the `max_load` to `0` seems to be necessary...  I've just found another stupid bug in `kobo` that lets the worker open all tasks that were returned in the list of openable tasks with the `ShutdownWorker` task.  I really hate `kobo`...

Sorry for making this even more confusing than it already is. :/

osh/hub/osh_xmlrpc/worker.py
lzaoral: That view is provided by `kobo` and filtering on the `method` task attribute is not implemented.

osh/hub/osh_xmlrpc/worker.py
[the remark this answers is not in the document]
lzaoral: Kobo PR with the fix: https://github.com/release-engineering/kobo/pull/255

osh/hub/scripts/osh-worker-manager
```diff
@@ -0,0 +1,228 @@
...
+    kobo_worker = models.Worker.objects.get(name=worker_name)
+    kobo_worker.delete()
+
+
+def workers_needed():
+    import kobo.hub.models
+    from django.conf import settings
+
+    # Tasks in opened or assigned state, skip counting subtasks
+    running_tasks = kobo.hub.models.Task.objects.running().filter(parent=None).count()
```
siteshwar, line 157: We need to skip `ShutdownWorker` tasks here.

osh/hub/scripts/osh-worker-manager
```diff
@@ -0,0 +1,228 @@
...
+
+    # Tasks waiting to be assigned, skip counting subtasks
```
siteshwar, line 159: We need to skip `ShutdownWorker` tasks here.

siteshwar:
> I find the use of `stdout`/`stderr` in `worker-manager` inconsistent. I think it should use `stdout` for output that is processed by the caller and `stderr` for diagnostic/progress output.

I tried to fix it now. Please check if I missed anything.

kdudka:
@siteshwar So in the end, only `--workers-needed` prints parseable output.  As I understand it, `--check-finished` provides the answer via exit status.  This is not entirely obvious and it should be described in the help string.

siteshwar:
@kdudka @lzaoral Can you take another look at this pull request?

osh/hub/scripts/osh-worker-manager
```diff
@@ -0,0 +1,244 @@
...
+    # TODO: Check for stale workers that have max_load above 0, but do not have arch
+    # or channel set. They would never be used.
+
+    # TODO: Shall we perform further health checks here? For example, if `osh-worker`
+    # service is running on the worker machine?
+    if kobo_worker.max_load > 0:
+        print(f"{worker_name} is new.", file=sys.stderr)
+        sys.exit(1)
+
+    if kobo_worker.running_tasks().count() > 0:
```
siteshwar, line 200: @lzaoral Would it be safer to use `task.is_finished()` (or some other) method here?

siteshwar: I added a check for disabled workers, so this question is irrelevant now.

osh/hub/scripts/osh-worker-manager
```diff
@@ -0,0 +1,248 @@
...
+        print(f"{worker_name} is running a task.", file=sys.stderr)
+        sys.exit(1)
+
+    if kobo_worker.enabled:
```
siteshwar, line 204: This change seems to work fine in my testing. E-mail notifications are sent and worker is deleted after it has been disabled.

osh/hub/osh_xmlrpc/worker.py
```diff
@@ -53,8 +56,15 @@
 
 @validate_worker
 def email_task_notification(request, task_id):
-    logger.debug('email_task_notification for %s', task_id)
-    return send_task_notification(request, task_id)
+    try:
+        logger.debug('email_task_notification for %s', task_id)
+        return send_task_notification(request, task_id)
+    finally:
+        if settings.ENABLE_SINGLE_USE_WORKERS:
```
siteshwar, line 63: TODO: Same check needs to be added to the `email_scan_notification` function.

kdudka: Unless you are going to resolve this TODO before merging, please insert a TODO comment at the place in the code where the check needs to be added.

osh/hub/scripts/osh-worker-manager
```diff
@@ -0,0 +1,253 @@
...
+    worker_conf = render_to_string("worker.conf.j2",
+                                   {"SINGLE_USE_WORKER_OSH_HUB_URL": settings.SINGLE_USE_WORKER_OSH_HUB_URL,
+                                    "OSH_WORKER_KEY": kobo_worker.worker_key})
+
+    # Create a temporary file and save the template
+    # Set the parameter to `delete_on_close=False` on python3.12 update
+    # https://docs.python.org/3.12/whatsnew/3.12.html#tempfile
+    # with tempfile.NamedTemporaryFile(delete=False) as temporary_file:
+    try:
+        temporary_file = tempfile.NamedTemporaryFile()
```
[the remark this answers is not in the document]
siteshwar: Please see the comment above. This feature is new in python-3.12 and we are using an older version of python in c9s.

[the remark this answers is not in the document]
kdudka: I do not get it.  The `with ... as` construction is not new in Python 3.12 and you call `temporary_file.close()` in the `finally` section anyway.

osh/hub/scripts/osh-worker-manager
```diff
@@ -0,0 +1,253 @@
...
+        temporary_file.write(worker_conf.strip().encode())
+        # TODO: File is closed, but not removed if `delete_on_close` is True.
```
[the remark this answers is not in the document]
siteshwar: It is for the time when we start using `delete_on_close` attribute with python-3.12.

osh/hub/osh_xmlrpc/worker.py
siteshwar: I tried to fix it now.

osh/hub/osh_xmlrpc/worker.py
```diff
@@ -76,11 +86,52 @@ def finish_task(request, task_id):
...
 
 @validate_worker
 def email_scan_notification(request, scan_id):
-    scan_notification_email(request, scan_id)
+    try:
+        logger.debug('email_scan_notification for %s', scan_id)
+        scan_notification_email(request, scan_id)
+    finally:
+        if settings.ENABLE_SINGLE_USE_WORKERS:
+            task = ScanBinding.objects.get(scan__id=scan_id).task
```
siteshwar, line 131: Does this line look correct? I do not know how to test it on local system.

osh/hub/scripts/osh-worker-manager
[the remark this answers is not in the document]
siteshwar: Ok, I get it now. Thanks!

osh/hub/scripts/osh-worker-manager
```diff
@@ -0,0 +1,253 @@
...
+        # This requires the hub to have ssh access to the new worker
+        scp_cmd = f"scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o ConnectTimeout=20 -i {settings.SINGLE_USE_WORKER_SSH_PRIVATE_KEY} {temporary_file.name} {settings.SINGLE_USE_WORKER_SSH_USER}@{kobo_worker.name}:worker.conf"
+        print(scp_cmd, file=sys.stderr)
+        scp_exit_status = subprocess.call(scp_cmd, shell=True)
+        if scp_exit_status != 0:
+            print(f"Failed to copy worker.conf to new worker {kobo_worker.name}", file=sys.stderr)
+            delete_kobo_worker(kobo_worker)
+            sys.exit(scp_exit_status)
+    finally:
+        temporary_file.close()
```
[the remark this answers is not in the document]
siteshwar: I have removed this code.

osh/hub/osh_xmlrpc/worker.py
kdudka: It does but I do not have a testing environment for this either.

kdudka (commented):
@siteshwar I am fine with merging this as it is but I can see a lot of unresolved comments by @lzaoral.  Please close the review comments that are resolved and comment on the unresolved ones.

osh/hub/settings_local.py
```diff
@@ -145,4 +145,21 @@
 # MOCK_AUTO_EXTRA_REPOS = {r'f\d+': '[repo]\n...'}
 MOCK_AUTO_EXTRA_REPOS = {}
 
+# If this setting is enabled, a worker is only used to perform a single task.
+ENABLE_SINGLE_USE_WORKERS = False
+# This number should be same as the `max` field in the `/etc/resallocserver/pools.yaml`
```
kdudka, line 150: Please insert a blank line to delimit the options.

osh/hub/scan/notify.py
```diff
@@ -38,8 +38,11 @@ def send_mail(message, subject, recipients, headers, bcc=None):
     headers["X-Application-ID"] = "OpenScanHub"
     headers["X-Hostname"] = socket.gethostname()
 
-    return EmailMessage(subject, message, from_addr, recipients, bcc=bcc,
-                        headers=headers, connection=connection).send()
+    logger.debug('Sending e-mail ... to %s', ", ".join(recipients))
```
kdudka, line 41: The three dots seem misplaced.

osh.spec
```diff
@@ -298,6 +309,11 @@ fi
 %postun hub
 %systemd_postun osh-{retention,stats}.{service,timer}
 
+%files worker-manager
+%defattr(-,root,apache,-)
```
kdudka, line 313: The use of `%defattr` seems undesired to me.  The binary is readable/executable by anyone, so it does not need to be in the `apache` group.  Same with the `/etc/osh/worker-manager` directory.  So I think that the `%defattr` line could/should be removed.

kdudka (approved):

kdudka:
@siteshwar I can see you rebased the pull request but the above review comments have not been addressed?

siteshwar:
> @siteshwar I can see you rebased the pull request but the above review comments have not been addressed?

I am deploying to Fedora infrastructure from this branch, so I was trying keep it up to date. I would get back to your comments probably next week. Thanks!

PR #237, opened 2024-03-08
containers: only set the worker container arch if the fixtures were loaded

lzaoral:
This fixes a `osh-hub` container crash when starting with the database snapshot restored.
```
+ python3 scripts/add-worker-arch.py
  File "scripts/add-worker-arch.py", line 28, in <module>
    main()
  File "scripts/add-worker-arch.py", line 19, in main
    worker = Worker.objects.get(name='localhost')
  File "/usr/lib/python3.6/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/django/db/models/query.py", line 437, in get
    self.model._meta.object_name
kobo.hub.models.DoesNotExist: Worker matching query does not exist.
```


kdudka (approved):
Looks good.  Thanks!

PR #238, opened 2024-03-08
client: fix the `--hub` option

lzaoral:
Previously, the code in OSH was not properly detecting the required support in `kobo`.

Fixes: 6d3e3cb3d57b0f4ca442ab919d2b6ee097032041 ("client: handle the --hub option globally")

### DIFF
--- a/osh/client/__init__.py
+++ b/osh/client/__init__.py
@@ -16,7 +16,7 @@ class OshCommand(ClientCommand):
 
         # For compatibility with older kobo releases
         sig = inspect.signature(self.set_hub)
-        if len(sig.parameters) == 4:
+        if 'hub' in sig.parameters:
             params.append(hub)
 
         self.set_hub(*params)


kdudka (changes requested):
The problem does not seem to be fully fixed.  The following command still silently ignores the `--hub` option:
```
% osh-cli watch-log --hub=https://covscan.lab.eng.brq2.redhat.com/osh/xmlrpc 467390
```
... despite the option is documented:
```
% osh-cli watch-log --help
Usage: osh-cli watch-log task_id

Options:
  -h, --help           show this help message and exit
  --username=USERNAME  specify user
  --password=PASSWORD  specify password
  --hub=HUB            specify URL of XML-RPC interface on hub
  --type=TYPE          Show log with this name, default is stdout.log
  --poll=POLL          Interval how often server should be polled for new info
                       (seconds >= 15)
  --nowait             Return after fetching current logfile, don't wait until
                       task finishes
```

lzaoral:
You have uncovered another bug, @kdudka!  The thing is that `watch-log` command does not use XML-RPC at all, so the option is silently ignored: https://github.com/release-engineering/kobo/blob/88237f21fce90b80135aeeaf20a602946c31586d/kobo/client/commands/cmd_watch_log.py#L60

Couple please try any command that is actually implemented in OSH?  That should be fixed by this PR.

lzaoral:
@kdudka FYI, https://github.com/release-engineering/kobo/pull/252 was merged.

kdudka (approved):
@lzaoral Perfect.  Thank you for fixing both the issues!

PR #239, opened 2024-03-18
client: catch network connectivity exceptions when verifying builds

rhyw:
Related: https://issues.redhat.com/browse/OSH-144

### DIFF
--- a/osh/client/commands/shortcuts.py
+++ b/osh/client/commands/shortcuts.py
@@ -32,7 +32,8 @@ def verify_build_exists(nvr, profile):
     try:
         # getBuild XML-RPC call is defined here: ./hub/kojihub.py:3206
         build = proxy_object.getBuild(nvr)
-    except koji.GenericError:
+    except Exception as e:  # noqa: B902
+        print('koji:', e, file=sys.stderr)
         return False
 
     if build is None:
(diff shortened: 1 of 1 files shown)


lzaoral (changes requested):
This seems to be quite overcomplicated.  What about just this instead?

```diff
diff --git a/osh/client/commands/shortcuts.py b/osh/client/commands/shortcuts.py
index 990f029e..739da189 100644
--- a/osh/client/commands/shortcuts.py
+++ b/osh/client/commands/shortcuts.py
@@ -24,15 +24,10 @@ def verify_build_exists(nvr, profile):
     """
     try:
         cfg = koji.read_config(profile)
-    except koji.ConfigurationError as e:
-        print('koji:', e, file=sys.stderr)
-        return False
-
-    proxy_object = koji.ClientSession(cfg['server'])
-    try:
-        # getBuild XML-RPC call is defined here: ./hub/kojihub.py:3206
+        proxy_object = koji.ClientSession(cfg['server'])
         build = proxy_object.getBuild(nvr)
-    except koji.GenericError:
+    except Exception as e:
+        print('koji:', e, file=sys.stderr)
         return False

     if build is None:

```

rhyw:
> This seems to be quite overcomplicated. What about just this instead?
> 
I tested, this also works. The reason I use "print_traceback" is to avoid the potential exception chains raised.

rhyw:
> > This seems to be quite overcomplicated. What about just this instead?
> 
> I tested, this also works. The reason I use "print_traceback" is to avoid the potential exception chains raised.
> +    except Exception as e:
> +        print('koji:', e, file=sys.stderr)

My bad, this should only print the str repr of the exception only. I'll give some tests before updating.

lzaoral (approved):

PR #240, opened 2024-03-18
hub: remove `task/et` view

lzaoral:
It has been broken for a very long time and there is no reason to give special spotlight to `ErrataDiffBuild` tasks from the upstream perspective.

Resolves: https://issues.redhat.com/browse/OSH-226


kdudka (approved):
Looks good.  Thanks!

PR #241, opened 2024-03-19
packaging: do not trigger `osh-{retention,stats}.service` on each update of the RPM packages

kdudka:
They will be triggered by the systemd timer according to the configured schedule.

Also do not restart `osh-{retention,stats}.service` on failure.  These are one-shot services.  If something fails, which should not happen under normal circumstances, they will be started again by the systemd timer later on (daily in the default configuration).

Fixes: https://github.com/openscanhub/openscanhub/issues/235


lzaoral (approved):
LGTM!

jamacku (approved):
LGTM

kdudka:
Thank you both for the quick review!

PR #242, opened 2024-03-20
hub: some minor improvements in find_tasks.py

rhyw:


lzaoral:
@rhyw Please, rebase the PR.

rhyw:
> @rhyw Please, rebase the PR.

Done

osh/hub/osh_xmlrpc/scan.py
```diff
@@ -237,30 +237,36 @@ def find_tasks(request, query):
...
-    comment = query.get('comment')
+    nvr = query.get('nvr')
     regex = query.get('regex')
+    comment = query.get('comment')
     states = query.get('states')
     latest = query.get('latest')
 
     result = []
-    tasks = None
     if nvr:
```
lzaoral, line 259: Off-topic:  This PR made me realise that the searching is broken/unintuitive when multiple options are specified, e.g.:
```bash
$ osh-cli find-tasks --nvr --comment test
```
is accepted but the comment in this function says that:

> Query is a dict (hash/map), it has to have exactly one of these keys:

This is not checked anywhere so CLI will still return some results...

@kdudka We should definitely fix this!

osh/hub/osh_xmlrpc/scan.py
```diff
@@ -237,30 +237,36 @@ def find_tasks(request, query):
...
     package_name = query.get('package_name')
-    comment = query.get('comment')
+    nvr = query.get('nvr')
     regex = query.get('regex')
+    comment = query.get('comment')
     states = query.get('states')
     latest = query.get('latest')
 
     result = []
-    tasks = None
```
lzaoral, line 255: nit: You can use `tasks = Task.objects.none()` here directly.  Then, the newly added `else` branch on L267 is redundant.  This will also make sure that the `tasks` variable will be always set to some instance of `QuerySet`.

osh/hub/osh_xmlrpc/scan.py
```diff
@@ -237,30 +237,36 @@ def find_tasks(request, query):
...
     elif regex:
         tasks = Task.objects.filter(label__regex=regex)
+    elif comment:
+        tasks = Task.objects.filter(comment__regex=comment)
+    else:
+        tasks = Task.objects.none()
+
     if states:
         tasks = tasks.filter(state__in=states)
     if tasks is not None:
```
lzaoral, line 272: This check is now redundant because `tasks` will never be set to `None`.

rhyw: updated

osh/hub/osh_xmlrpc/scan.py
rhyw: updated.

osh/hub/osh_xmlrpc/scan.py
rhyw: filed an issue so we get somewhere to track: https://github.com/openscanhub/openscanhub/issues/268

lzaoral: Thank you!

lzaoral (approved):
Thank you, @rhyw!

kdudka (approved):

Issue #133, opened 2023-09-06
hub: repurpose the `results` field for `Task` result summary


lzaoral:
Right now, all successful tasks (and a lot of failed ones) have an empty results field.  It could be useful to repurpose this field to show the results statistics (e.g. the output of `csgrep --mode=stat`) for successful tasks and a short summary of a failure for the failed tasks.

PR #246, opened 2024-03-21
hub: embed results summary in `Task.result` field

lzaoral:
Resolves: https://github.com/openscanhub/openscanhub/issues/133

### DIFF
--- a/osh/hub/waiving/results_loader.py
+++ b/osh/hub/waiving/results_loader.py
@@ -45,10 +45,22 @@ class TaskResultsProcessor:
         if task_has_results(self.target_task):
             logger.info("Results are already unpacked for task %s", self.target_task)
             return
-        else:
-            logger.debug('Unpacking %s', tb_path)
-            rex = ResultsExtractor(tb_path, output_dir=self.target_task_dir, unpack_in_temp=False)
-            rex.extract_tarball(self.exclude_dirs)
+
+        logger.debug('Unpacking %s', tb_path)
+        rex = ResultsExtractor(tb_path, output_dir=self.target_task_dir, unpack_in_temp=False)
+        rex.extract_tarball(self.exclude_dirs)
+
+        try:
+            with open(self.target_paths.get_txt_summary()) as f:
+                # read at most 256 KiB
+                self.target_task.result = f.read(256 * 1024).rstrip()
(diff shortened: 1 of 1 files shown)


osh/hub/waiving/results_loader.py
```diff
@@ -45,10 +45,18 @@ def unpack_results(self):
...
-            rex = ResultsExtractor(tb_path, output_dir=self.target_task_dir, unpack_in_temp=False)
-            rex.extract_tarball(self.exclude_dirs)
+
+        logger.debug('Unpacking %s', tb_path)
+        rex = ResultsExtractor(tb_path, output_dir=self.target_task_dir, unpack_in_temp=False)
+        rex.extract_tarball(self.exclude_dirs)
+
+        try:
+            with open(self.target_paths.get_txt_summary()) as f:
+                self.target_task.result = f.read().rstrip()
```
kdudka, line 55: Is there any size limit in the model?  If not we should enforce a size limit here, so that we do not store 1 GiB summary into the database if a scan goes wild.

lzaoral: No, there is not: https://github.com/release-engineering/kobo/blob/9af65943b183e2904b47855cbc820b8dda2d5643/kobo/hub/models.py#L561

How many characters at most do you suggest to store, e.g. `5000`?

kdudka: We have valid summary files with ~20 KiB on the internal OSH instance, e.g. https://cov01.lab.eng.brq2.redhat.com/covscanhub/task/486761/log/kernel-5.14.0-427.5.1.el9_4/scan-results-all-summary.txt

I would set the limit to catch obvious mistakes only.  What about 256 KiB?

kdudka: Alternatively, we can keep it as a preview only.  If the summary is truncated, we can append a warning with a reference to full results, like we do for `stdout.log`.

lzaoral: Addressed in 90a364bf995abf5ed3f16dbbb286e66252e9d163.

rhyw:
LGTM

kdudka (approved):
Works as expected.  Thanks for the update!

PR #247, opened 2024-03-21
hub: enable reading of secrets in `settings_local.py`

lzaoral:
hub: enable reading of secrets in `settings_local.py`

Resolves: https://issues.redhat.com/browse/OSH-387

### DIFF
--- a/osh/hub/other/settings_util.py
+++ b/osh/hub/other/settings_util.py
@@ -1,0 +1,12 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+# SPDX-FileCopyrightText: Copyright contributors to the OpenScanHub project.
+
+import os
+
+
+def get_secret(name, directory):
+    try:
+        with open(os.path.join(directory, name)) as f:
+            return f.read().strip()
+    except OSError:
+        return None
--- a/osh/hub/settings.py
+++ b/osh/hub/settings.py
@@ -179,6 +179,13 @@ LOGIN_EXEMPT_URLS = ['.*xmlrpc/.*']
 LOGIN_REDIRECT_URL = 'index'
 LOGOUT_REDIRECT_URL = 'index'
 
+# Denote whether the access to user list/detail view is restricted
+# Possible values:
+# "" (empty string) = Anonymous access (default)
+# "authenticated" = Authenticated users
+# "staff" = Staff (admin) users only
+USERS_ACL_PERMISSION = "staff"
+
 VALID_TASK_LOG_EXTENSIONS = ['.log', '.ini', '.err', '.out', '.js', '.txt']
 
 # override default values with custom ones from local settings
@@ -199,12 +206,6 @@ def _get_secret(name):
 BZ_API_KEY = _get_secret('bugzilla_secret')
 JIRA_API_KEY = _get_secret('jira_secret')
 
-# Denote whether the access to user list/detail view is restricted
-# Possible values:
-# "" (empty string) = Anonymous access (default)
-# "authenticated" = Authenticated users
-# "staff" = Staff (admin) users only
-USERS_ACL_PERMISSION = "staff"
 
 # read the real SECRET_KEY from SECRET_KEY_FILE if availble
 try:
--- a/osh/hub/settings.py
+++ b/osh/hub/settings.py
@@ -16,8 +16,8 @@ PROJECT_DIR = os.path.dirname(__file__)
 
 URL_PREFIX = "/osh"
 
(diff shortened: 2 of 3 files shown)


osh/hub/settings.py
```diff
@@ -188,28 +192,15 @@
...
+JIRA_API_KEY = get_secret('jira_secret', SECRETS_DIR)
 
-# read the real SECRET_KEY from SECRET_KEY_FILE if availble
-try:
-    with open(SECRET_KEY_FILE) as f:
-        key = f.readline()
-        SECRET_KEY = key.strip()
-except OSError:
-    pass
+# Secret key ((will be overridden by the content of SECRET_KEY_FILE if available)
```
kdudka, line 205: `((` -> `(`

osh/hub/settings.py
```diff
@@ -188,28 +192,15 @@
...
-    except OSError:
-        return None
-
+###############################################################################
+# Secrets
+###############################################################################
 
-BZ_API_KEY = _get_secret('bugzilla_secret')
-JIRA_API_KEY = _get_secret('jira_secret')
+from osh.hub.other.settings_util import get_secret  # noqa: E402
```
kdudka, line 199: Does it still need to be imported here?  The access to `SECRETS_DIR` from the function is removed in d5372cae14172eedc45f6aa2bea0f9c648229f64.

osh/hub/settings_local.py
```diff
@@ -25,6 +25,7 @@
         'ENGINE': 'django.db.backends.postgresql',
         'NAME': 'openscanhub',
         'USER': 'openscanhub',
+        # Use osh.hub.other.settings_util.get_secret to load the secret from file.
         'PASSWORD': 'velryba',
```
kdudka, line 29: Could we use something like this?
```py
       'PASSWORD': get_secret('db_password', SECRET_KEY_DIR) or 'velryba',
```

This way a user could protect the development database with a password without preventing `settings_local.py` from being updated in the future.

osh/hub/settings.py
lzaoral: Yes, otherwise the `get_secret` function would be undefined and Python Interpreter would raise `NameError`.

lzaoral: Well, technically, the symbol might get imported transitively from `settings_local.py` but that is not guaranteed.

osh/hub/settings.py
lzaoral: Addressed in ff868222d7d675213c74a4dfffabd63d0a8d4f84.

osh/hub/settings_local.py
lzaoral: Addressed in 1cc549a8530937a6c8ee8971ae174aac21b6b93a.

osh/hub/settings.py
kdudka: I understand that `get_secert()` needs to be imported.  I am just asking why it cannot be imported at the top of this file without using `# noqa: E402`.

lzaoral: Sorry, I have missed this reply.  I've just aggregated all secret-related stuff to one place.  Addressed in a1d083c25b781366f45fedfe6097aedb3956c385.

osh/hub/other/settings_util.py
```diff
@@ -0,0 +1,12 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+# SPDX-FileCopyrightText: Copyright contributors to the OpenScanHub project.
+
+import os
+
+
+def get_secret(name, directory):
+    try:
+        with open(os.path.join(directory, name)) as f:
```
hanchuntao, line 9: `encoding="utf-8"` ?

lzaoral: Thank you for the suggestion but `utf-8` is the default on all reasonably modern Unix-like systems.  The only major OS which still defaults to something else is Windows which is unsupported by OSH anyway.

hanchuntao: Yes. From the buildins : `In text mode, if encoding is not specified the encoding used is platform dependent: locale.getpreferredencoding(False) is called to get the current locale encoding.`
```
[root@osh-test-worker-001 ~]# python
Python 3.9.18 (main, Jan  4 2024, 00:00:00) 
[GCC 11.4.1 20230605 (Red Hat 11.4.1-2)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import locale
>>> print(locale.getpreferredencoding(False))
UTF-8
```

hanchuntao (approved):

PR #249, opened 2024-04-03
worker: do not use `cspodman --force-global-cleanup-on-exit`

kdudka:
It was a potentially harmful interim solution, which is no longer needed.
The option is now ignored by cspodman anyway:
```
% cspodman --force-global-cleanup-on-exit ... 
cspodman: warning: --force-global-cleanup-on-exit is not implemented
[...]
``` 

Related: https://issues.redhat.com/browse/OSH-473

### DIFF
--- a/osh/worker/csmock_runner.py
+++ b/osh/worker/csmock_runner.py
@@ -115,8 +115,6 @@ class CsmockRunner:
 
         if profile == "cspodman":
             cmd = "cspodman"
-            # XXX: this destroys all containers we have access to, related or not
-            cmd += " --force-global-cleanup-on-exit"
         else:
             cmd = "csmock"
             if profile_url:


kdudka:
I need to put this change on hold temporarily until we resolve SELinux issues that we hit on production OSH workers caused by the fix that eliminates the need for `cspodman --force-global-cleanup-on-exit`.

kdudka:
The issue mentioned in the previous comment has been resolved via https://issues.redhat.com/browse/OSH-606.

I think we can safely merge this now...

osh/worker/csmock_runner.py
```diff
@@ -119,8 +119,6 @@ def analyze(self, analyzers, srpm_path, profile=None, su_user=None, additional_a
 
         if profile == "cspodman":
             cmd = "cspodman"
-            # XXX: this destroys all containers we have access to, related or not
-            cmd += " --force-global-cleanup-on-exit"
```
hanchuntao, line 123: If we don't need to use it, can we remove all of the code, such as:
https://gitlab.cee.redhat.com/osh/cspodman/-/blame/main/py/cspodman#L62
https://github.com/csutils/csdiff/blob/main/tests/csgrep/0117-csgrep-set-imp-level-stdin.txt#L19
https://github.com/csutils/csdiff/blob/main/tests/csgrep/0117-csgrep-set-imp-level-stdout.txt#L19

kdudka: The option has existed in `cspodman` since 2020.  I think it should go away through a deprecation process.  The code that changed behavior of `cspodman` has already been removed.  Removing the warning right now is not going to help.  It would only break OSH if the changes were applied in wrong order.

The files in `csgrep` are test coverage taken from old scans.  I do not think we need to change scan properties retroactively.

lzaoral: I agree, there is no reason to remove it from test files.

lzaoral (approved):

PR #254, opened 2024-04-08
hub: prepend a `/` in the template path override

siteshwar:


siteshwar:
I discovered this issue while testing changes for https://github.com/openscanhub/fedora-infra/issues/31.

The [staging](https://openscanhub.stg.fedoraproject.org/) deployment still uses the default footer, even though `/etc/osh/hub/templates/footer.html` file is present with expected content:

```
$ cat /etc/osh/hub/templates/footer.html
Issues can be reported to the <a href="https://lists.fedoraproject.org/archives/list/openscanhub@lists.fedoraproject.org/">mailing list</a>.
```

From the `fedora-osh-hub` container:

```
>>> import os
>>> os.path.join("etc", "osh", "hub", "templates")
'etc/osh/hub/templates'
```

It is missing a `/` at the beginning, so overrides would not work, as current directory is not root directory.

osh/hub/settings.py
```diff
@@ -56,7 +56,7 @@
         'BACKEND': 'django.template.backends.django.DjangoTemplates',
         'DIRS': [
             # `/etc/osh/hub/templates` overrides `index-content.html`
-            os.path.join("etc", "osh", "hub", "templates"),
+            os.path.join(os.path.sep, "etc", "osh", "hub", "templates"),
```
kdudka, line 59: I guess this worked only because `systemd` runs `httpd` with `cwd` set to `/`.  Anyway, I do not think that `os.path.join()` adds here any value.  Cannot we just replace it with `"/etc/osh/hub/templates"`?

It will be much easier to read...

rhyw (approved):

kdudka (approved):

osh/hub/settings.py
```diff
@@ -56,7 +56,7 @@
         'BACKEND': 'django.template.backends.django.DjangoTemplates',
         'DIRS': [
             # `/etc/osh/hub/templates` overrides `index-content.html`
-            os.path.join("etc", "osh", "hub", "templates"),
+            os.path.join("/etc/osh/hub/templates"),
```
lzaoral, line 59: TBH, calling `os.path.join` with a single argument is completely useless.

hanchuntao: if we want to use L59, why not use `"/etc/osh/hub/templates"` directly, or use `os.path.join("/", "etc", "osh", "hub", "templates")`

siteshwar: https://github.com/openscanhub/openscanhub/pull/264/files

Issue #253, opened 2024-04-05
hub: domain name for notification e-mails should be configurable


siteshwar:
The `email` field for users is not autogenerated and domain name is [hard coded](https://github.com/openscanhub/openscanhub/blob/f5ccf9619c07586f1c0b6ba8cdcc5b0ba81a3805/osh/hub/scan/notify.py#L60) in the sources. This should be configurable through settings.

PR #256, opened 2024-04-15
hub: make domain name for sending notifications

siteshwar:
... to the user configurable.

Resolves: https://github.com/openscanhub/openscanhub/issues/253

### DIFF
--- a/osh/hub/scan/notify.py
+++ b/osh/hub/scan/notify.py
@@ -56,8 +56,7 @@ def get_recipient(user):
     if user.email:
         return user.email
 
-    # XXX: hardcoded
-    return user.username + "@redhat.com"
+    return f"{user.username}@{settings.DEFAULT_EMAIL_DOMAIN}"
 
 
 def generate_stats(task, diff_task=False, with_defects_in_patches=False):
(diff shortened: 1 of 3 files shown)


osh/hub/scan/notify.py
```diff
@@ -56,8 +56,7 @@ def get_recipient(user):
     if user.email:
         return user.email
 
-    # XXX: hardcoded
-    return user.username + "@redhat.com"
+    return "@".join([user.username, settings.DEFAULT_EMAIL_DOMAIN])
```
kdudka, line 59: I find this unnecessarily cryptic.  Could we please write this as `user.username + "@" + settings.DEFAULT_EMAIL_DOMAIN` or `f"{user.username}@{settings.DEFAULT_EMAIL_DOMAIN}"`?

kdudka (changes requested):
I would move the previously hardcoded value to `settings.py` to make the change backward compatible and to avoid unnecessary tracebacks after updates.

osh/hub/settings_local.py
```diff
@@ -130,4 +130,9 @@
 # E-mail address where notifications about failed/cancelled tasks are sent
 DEVEL_EMAIL_ADDRESS = "no-reply@example.org"
 
+# Default e-mail domain name for sending notifications to the user.
+# It is used when `email` field in the user model is not set.
+# For example: username@example.org
```
kdudka, line 135: I find the `For example: username@example.org` line redundant and confusing.

kdudka (approved):

PR #259, opened 2024-04-18
spec: tweak spec file based on Fedora review

siteshwar:
Related: https://bugzilla.redhat.com/show_bug.cgi?id=2275304

### DIFF
--- a/osh.spec
+++ b/osh.spec
@@ -6,6 +6,7 @@ Release:        1%{?dist}
 License:        GPL-3.0-or-later
 Summary:        Static and Dynamic Analysis as a Service
 URL:            https://github.com/openscanhub/openscanhub/
+# Sources are available at https://github.com/openscanhub/openscanhub/tags
 Source:         https://github.com/openscanhub/openscanhub/archive/refs/tags/%{name}-%{version}.tar.gz
 BuildArch:      noarch
 
@@ -29,7 +30,7 @@ BuildRequires:  systemd-rpm-macros
 
 
 %description
-OpenScanHub is a service for static and dynamic analysis.
+OpenScanHub is a service for static and dynamic analysis of RPM packages.
 It consists of central hub, workers and cli client.
 
 
@@ -47,10 +48,10 @@ Provides: osh-cli = %{version}-%{release}
 # This is kept here for backward compatibility with old package name
 Provides: covscan-client = %{version}
 Obsoletes: covscan-client < %{version}
-Conflicts: covscan-client < %{version}
 
 %description client
-OpenScanHub CLI client
+OpenScanHub CLI client.
+It is used to submit, query and download scans.
 
 
 %package common
@@ -75,7 +76,8 @@ Recommends: osh-worker-conf
 Obsoletes: covscan-worker < %{version}
 
 %description worker
-OpenScanHub worker
+OpenScanHub worker.
+It is used to process tasks submitted to the hub.
 
 %package hub
 Summary: OpenScanHub xml-rpc interface and web application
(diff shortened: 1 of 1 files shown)


siteshwar:
Shall I bump up the version number on this change?

osh.spec
```diff
@@ -6,6 +6,7 @@ Release:        1%{?dist}
 License:        GPL-3.0-or-later
 Summary:        Static and Dynamic Analysis as a Service
 URL:            https://github.com/openscanhub/openscanhub/
+# Sources are available at https://github.com/openscanhub/openscanhub/tags
```
siteshwar, line 9: The full URL is removed by PackIt, so I am leaving a comment here.

kdudka:
> Shall I bump up the version number on this change?

No.

kdudka (changes requested):
Per https://bugzilla.redhat.com/show_bug.cgi?id=2275304#c2 you should also remove `Conflicts: covscan-client < %{version}`.

The comments about the description were IMO not related to the top-level `osh` package, which is non-existent for end users, but about all the subpackages.

osh.spec
```diff
@@ -31,6 +32,7 @@ BuildRequires:  systemd-rpm-macros
 %description
 OpenScanHub is a service for static and dynamic analysis.
 It consists of central hub, workers and cli client.
+It runs various static (and dynamic) analyzers on RPM packages.
```
kdudka, line 35: I find the description a bit clumsy.  The first and last sentence duplicate each other.  Cannot you just append ` of RPM packages` to the first sentence?

osh.spec
```diff
@@ -75,7 +77,8 @@ Recommends: osh-worker-conf
 Obsoletes: covscan-worker < %{version}
 
 %description worker
-OpenScanHub worker
+OpenScanHub worker.
+It is used to perform scans.
```
kdudka, line 81: I am not sure if this is an improvement really.  I would say that all the subpackages are used "to perform scans".  I would reuse the description we already have in `kobo`: https://src.fedoraproject.org/rpms/kobo/blob/3b18f70fbe4809a4aa7be89b6c93b3d716738b60/f/kobo.spec#_73

osh.spec
```diff
@@ -47,10 +49,10 @@ Provides: osh-cli = %{version}-%{release}
 # This is kept here for backward compatibility with old package name
 Provides: covscan-client = %{version}
 Obsoletes: covscan-client < %{version}
-Conflicts: covscan-client < %{version}
 
 %description client
-OpenScanHub CLI client
+OpenScanHub CLI client.
+It is used to submit, query and download builds.
```
kdudka, line 55: `builds` -> `scans`

kdudka (changes requested):

kdudka (approved):

osh.spec
```diff
@@ -232,7 +231,6 @@ fi
 %attr(640,root,root) %config(noreplace) %{_sysconfdir}/osh/worker.conf
 
 %files hub
-%defattr(-,root,apache,-)
```
xsuchy, line 235: If you remove this then files will be owned by root:root.
I do NOT know what should be ownership of the files. It just seemed suspicious to me that all files in this subpackage are owned by root:apache.

siteshwar: Tests in [Testing Farm](https://artifacts.dev.testing-farm.io/1b0bc17a-a514-46d3-8aab-7c9d051b6350/) passed, so I believe these changes should not break existing deployments.

kdudka: @xsuchy I was double checking this before the approval.  The files that need to be in the `apache` group use `%attr` explicitly.  Everything else is just Python scripts and configuration files, where the group can be set to `root`.

PR #269, opened 2024-04-30
hub: add automatic mock config detection for container scans

lzaoral:
$TITLE.


kdudka (approved):
It works as expected.  Thanks!

PR #272, opened 2024-05-07
spec: some more fixes based on Fedora review

siteshwar:
Related: https://bugzilla.redhat.com/show_bug.cgi?id=2275304

### DIFF
--- a/osh.spec
+++ b/osh.spec
@@ -283,7 +283,12 @@ print(get_random_secret_key())" > %{_sharedstatedir}/osh/hub/secret_key
 fi
 
 # this only takes an effect if PostgreSQL is running and the database exists
-pg_isready -h localhost && %{python3_sitelib}/osh/hub/manage.py migrate
+if pg_isready -h localhost; then
+    # run `manage.py` as the `apache` user to improve security and to prevent
+    # the Python interpreter from creating an unowned byte-compiled module for
+    # `settings_local.py`
+    runuser -u apache -- %{python3_sitelib}/osh/hub/manage.py migrate
+fi
 
 %systemd_post osh-{retention,stats}.{service,timer}
 
--- a/osh.spec
+++ b/osh.spec
@@ -188,6 +188,10 @@ rm -rf %{buildroot}%{python3_sitelib}/scripts
 # install example httpd config
 install -D {osh/hub,%{buildroot}%{_sysconfdir}/httpd/conf.d}/osh-hub-httpd.conf
 
+# keep configuration in `/etc` so that it can be overridden if /usr is read-only
+mv %{buildroot}%{python3_sitelib}/osh/hub/settings_local.py %{buildroot}%{_sysconfdir}/osh/hub
+ln -s %{_sysconfdir}/osh/hub/settings_local.py %{buildroot}%{python3_sitelib}/osh/hub/settings_local.py
+
 %files client
 %attr(755,root,root) %{_bindir}/osh-cli
 %{_bindir}/covscan
@@ -241,13 +245,15 @@ fi
 %{_sbindir}/osh-retention
 %{_sbindir}/osh-stats
 %{_sysconfdir}/osh/hub
-%exclude %{python3_sitelib}/osh/hub/scripts/osh-xmlrpc-client.py
-%exclude %{python3_sitelib}/osh/hub/scripts/umb-emit.py
 %{python3_sitelib}/osh/hub
 %{_unitdir}/osh-retention.*
 %{_unitdir}/osh-stats.*
+%exclude %{python3_sitelib}/osh/hub/scripts/osh-xmlrpc-client.py*
+%exclude %{python3_sitelib}/osh/hub/scripts/umb-emit.py*
 %exclude %{python3_sitelib}/osh/hub/settings_local.py*
+%exclude %{python3_sitelib}/osh/hub/settings_local.ci.py*
 %exclude %{python3_sitelib}/osh/hub/__pycache__/settings_local.*
+%exclude %{_sysconfdir}/osh/hub/settings_local.py*
 %dir %{_localstatedir}/log/osh
 # These files should be readable and writable by respective groups.
 %dir %attr(775,root,apache) %{_localstatedir}/log/osh/hub
@@ -289,8 +295,8 @@ pg_isready -h localhost && %{python3_sitelib}/osh/hub/manage.py migrate
 
 %files hub-conf-devel
 %defattr(644,root,apache)
-%config(noreplace) %{python3_sitelib}/osh/hub/settings_local.py
-%config(noreplace) %{python3_sitelib}/osh/hub/__pycache__/settings_local*.pyc
+%{python3_sitelib}/osh/hub/settings_local.py
+%config(noreplace) %{_sysconfdir}/osh/hub/settings_local.py
 %config(noreplace) %{_sysconfdir}/httpd/conf.d/osh-hub-httpd.conf
 %ghost %attr(640,root,apache) %{_sysconfdir}/osh/hub/secrets/db_password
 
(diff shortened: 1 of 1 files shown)


osh.spec
```diff
@@ -241,13 +248,15 @@ fi
...
-%exclude %{python3_sitelib}/osh/hub/scripts/osh-xmlrpc-client.py
-%exclude %{python3_sitelib}/osh/hub/scripts/umb-emit.py
+%exclude %{python3_sitelib}/osh/hub/scripts/osh-xmlrpc-client.py*
+%exclude %{python3_sitelib}/osh/hub/scripts/umb-emit.py*
 %{python3_sitelib}/osh/hub
 %{_unitdir}/osh-retention.*
 %{_unitdir}/osh-stats.*
 %exclude %{python3_sitelib}/osh/hub/settings_local.py*
 %exclude %{python3_sitelib}/osh/hub/__pycache__/settings_local.*
+%exclude %{python3_sitelib}/osh/hub/__pycache__/settings.*
```
siteshwar, line 258: Shall this cache be kept? Where/how it should be moved?

osh.spec
```diff
@@ -188,6 +188,13 @@ rm -rf %{buildroot}%{python3_sitelib}/scripts
 # install example httpd config
 install -D {osh/hub,%{buildroot}%{_sysconfdir}/httpd/conf.d}/osh-hub-httpd.conf
 
+install -D {osh/hub,%{buildroot}%{_sysconfdir}/osh/hub}/settings.py
+install -D {osh/hub,%{buildroot}%{_sysconfdir}/osh/hub}/settings_local.ci.py
+
+rm -f %{buildroot}%{python3_sitelib}/osh/hub/settings.py
+rm -f %{buildroot}%{python3_sitelib}/osh/hub/settings_local.ci.py
```
kdudka, line 195: I would simply use `mv` to move the existing files.

osh.spec
kdudka: This glob is a superset of the above glob, isn't it?  None if them is needed I think.  `settings.py` is installed by `osh-hub` and the other files should not byte-compiled at this location at all.

osh.spec
```diff
@@ -276,6 +285,9 @@ print(get_random_secret_key())" > %{_sharedstatedir}/osh/hub/secret_key
     chgrp apache %{_sharedstatedir}/osh/hub/secret_key
 fi
 
+ln -s %{_sysconfdir}/osh/hub/settings.py %{python3_sitelib}/osh/hub/settings.py
+ln -s %{_sysconfdir}/osh/hub/settings_local.ci.py %{python3_sitelib}/osh/hub/settings_local.ci.py
```
kdudka, line 289: The symlinks should be created in `%install` so that they are managed by `rpm`.

kdudka (changes requested):

osh.spec
```diff
@@ -188,6 +188,16 @@ rm -rf %{buildroot}%{python3_sitelib}/scripts
 # install example httpd config
 install -D {osh/hub,%{buildroot}%{_sysconfdir}/httpd/conf.d}/osh-hub-httpd.conf
 
+# install -D {osh/hub,%{buildroot}%{_sysconfdir}/osh/hub}/settings.py
+# install -D {osh/hub,%{buildroot}%{_sysconfdir}/osh/hub}/settings_local.ci.py
+
+mv %{buildroot}%{python3_sitelib}/osh/hub/settings.py %{buildroot}%{_sysconfdir}/osh/hub
+mv %{buildroot}%{python3_sitelib}/osh/hub/settings_local.ci.py %{buildroot}%{_sysconfdir}/osh/hub/
```
siteshwar, line 195: On a second look, we may not need this file at all in the rpm package, as the CI uses it [directly](https://github.com/openscanhub/openscanhub/blob/bd8475834908a317f552cb9321e24b2da46721b1/tests/ci.sh#L12) from git repository.

siteshwar:
@xsuchy Would it be possible for you to take a look at it?

kdudka (changes requested):
The commit message sounds too vague to me.  Please reword it such that:
1. The commit title summarizes the changes.
2. The commit description explains **why** the changes are made (without the need to click any external links).

xsuchy:
The file `%{_sysconfdir}/osh/hub/settings.py` is now owned by -hub subpackage (and not marked as %config) while you want it to be owned by hub-conf-devel subpackage.

kdudka:
@xsuchy  This is expected.  `%{_sysconfdir}/osh/hub/settings.py ` should be owned by `osh-hub`.  But your comment made me realize that the proposed change is wrong.  We need to move and symlink `settings_local.py` rather than `settings.py`.

osh.spec
```diff
@@ -241,13 +244,16 @@ fi
...
+%exclude %{python3_sitelib}/osh/hub/scripts/osh-xmlrpc-client.py*
+%exclude %{python3_sitelib}/osh/hub/scripts/umb-emit.py*
 %{python3_sitelib}/osh/hub
 %{_unitdir}/osh-retention.*
 %{_unitdir}/osh-stats.*
 %exclude %{python3_sitelib}/osh/hub/settings_local.py*
+%exclude %{python3_sitelib}/osh/hub/settings_local.ci.py*
+%exclude %{_sysconfdir}/osh/hub/settings_local.py*
 %exclude %{python3_sitelib}/osh/hub/__pycache__/settings_local.*
+%exclude %{python3_sitelib}/osh/hub/__pycache__/settings.*
```
kdudka, line 256: I think this line was added by mistake.  `settings.py` is still owned by `osh-hub` and so should be the corresponding byte-code.

osh.spec
```diff
@@ -289,7 +295,8 @@ pg_isready -h localhost && %{python3_sitelib}/osh/hub/manage.py migrate
 
 %files hub-conf-devel
 %defattr(644,root,apache)
-%config(noreplace) %{python3_sitelib}/osh/hub/settings_local.py
+%{python3_sitelib}/osh/hub/settings_local.py
+%config(noreplace) %{_sysconfdir}/osh/hub/settings_local.py
 %config(noreplace) %{python3_sitelib}/osh/hub/__pycache__/settings_local*.pyc
```
kdudka, line 300: We do not want to include the byte-compiled content of `settings_local.py` into the package because the corresponding Python code is moved to `/etc/osh/hub` and can be modified by the admin.

kdudka (changes requested):

osh.spec
```diff
@@ -187,6 +187,9 @@ rm -rf %{buildroot}%{python3_sitelib}/scripts
 
 # install example httpd config
 install -D {osh/hub,%{buildroot}%{_sysconfdir}/httpd/conf.d}/osh-hub-httpd.conf
+# configuration files should be kept under `/etc`
+mv %{buildroot}%{python3_sitelib}/osh/hub/settings_local.py %{buildroot}%{_sysconfdir}/osh/hub
+ln -s %{_sysconfdir}/osh/hub/settings_local.py %{buildroot}%{python3_sitelib}/osh/hub/settings_local.py
```
kdudka, line 192: When I execute `/usr/lib/python3.12/site-packages/osh/hub/manage.py` as `root` on my Fedora 39 VM, the Python interpreter creates `/usr/lib/python3.12/site-packages/osh/hub/__pycache__/settings_local.cpython-312.pyc`, which is not owned by any RPM package.  @xsuchy Is this expected/acceptable?

kdudka: We can modify `%post` to run `manage.py` as the `apache` user.  Still if someone runs the script(s) as `root` manually an unowned file will be created in `%{buildroot}%{python3_sitelib}`.  I take you answer as "this is acceptable packaing-wise".

xsuchy: Right :)

kdudka: Fixed in de4295305b8aeccdb50b4100315ba08d9b571b61.

kdudka:
@siteshwar I have pushed a fixup to address https://github.com/openscanhub/openscanhub/pull/272#discussion_r1598522603 and two follow-up commits.  Could you please have a look?

kdudka (approved):

PR #274, opened 2024-05-09
hub/client: order tasks by id in find-tasks

rhyw:
Related: https://issues.redhat.com/browse/OSH-573

### DIFF
--- a/osh/hub/osh_xmlrpc/scan.py
+++ b/osh/hub/osh_xmlrpc/scan.py
@@ -269,7 +269,7 @@ def find_tasks(request, query):
     if states:
         tasks = tasks.filter(state__in=states)
 
-    result = tasks.order_by("-dt_finished").values_list("id", flat=True)
+    result = tasks.order_by("-id").values_list("id", flat=True)
     # truncate the result in case `latest` is specified, this reduces the amount of
     # data transferred over the network
     if latest:


osh/hub/osh_xmlrpc/scan.py
```diff
@@ -269,7 +269,7 @@ def find_tasks(request, query):
     if states:
         tasks = tasks.filter(state__in=states)
 
-    result = tasks.order_by("-dt_finished").values_list("id", flat=True)
```
lzaoral, line 272: Thank for the quick fix, @rhyw!

Are we sure that the assignment of the `id` value is monotonic?  In other words, are we sure that the DB engine will not reuse previously deleted Task `id`s for new Tasks?  It may be safer to sort on `dt_created` instead.

@kdudka WDYT?

kdudka: @lzaoral I do not think we have ever deleted any tasks in OSH.  The retention policy deletes logs (scan results) associated with tasks but there is no code that would delete the task objects themselves.  I would prefer to sort by task ID because that is a plain integer.  Sorting by time would cause problems in case date/time was incorrectly set on OSH machines etc.

lzaoral: Ok, but in that case, we should at least add a comment about this behaviour when tasks are deleted.

rhyw: By default the ID field is auto-incremental thus old IDs won't be re-used. I don't see any customization in kobo itself, thus I think it won't be a problem even if old tasks are deleted.

lzaoral: TIL, @rhyw!  Ref: https://docs.djangoproject.com/en/5.0/topics/db/models/#automatic-primary-key-fields

lzaoral (approved):

kdudka (approved):
Looks good.  Thank you both!

PR #278, opened 2024-05-23
mock-build to support --git-url option for scratch build

rhyw:
Resolves: https://issues.redhat.com/browse/OSH-592

### DIFF
--- a/osh/client/commands/cmd_build.py
+++ b/osh/client/commands/cmd_build.py
@@ -58,6 +58,7 @@ class Base_Build(OshCommand):
         config = kwargs.get(prefix + "config")
         nvr = kwargs.get(prefix + "nvr")
         srpm = kwargs.get(prefix + "srpm")
+        dist_git_url = kwargs.get(prefix + "git_url")
         tarball_build_script = kwargs.get(prefix + "tarball_build_script")
 
         if not config:
@@ -83,6 +84,8 @@ is not even one in your user configuration file \
             if result is not None:
                 self.parser.error(result)
             options[prefix + "brew_build"] = nvr
+        elif dist_git_url is not None:
+            options[prefix + "dist_git_url"] = dist_git_url
         else:
             if not os.path.exists(srpm):
                 self.parser.error(f"file does not exist: {srpm}")
--- a/osh/client/commands/cmd_build.py
+++ b/osh/client/commands/cmd_build.py
@@ -99,6 +99,17 @@ is not even one in your user configuration file \
 
         return options
 
+    def _process_srpm_option(self, args, kwargs):
+        """Validate the srpm option and update kwargs in place."""
+        if not args:
+            # do nothing if no args are provided
+            return
+
+        if len(args) != 1:
+            self.parser.error("please specify exactly one SRPM")
+
+        kwargs["srpm"] = os.path.abspath(os.path.expanduser(args[0]))
+
     def prepare_task_options(self, args, kwargs):
         # optparser output is passed via args (args) and kwargs (opts)
         analyzers = kwargs.get('analyzers', '')
--- a/osh/client/commands/cmd_mock_build.py
+++ b/osh/client/commands/cmd_mock_build.py
@@ -1,12 +1,12 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 # SPDX-FileCopyrightText: Copyright contributors to the OpenScanHub project.
 
-from osh.client.commands.cmd_diff_build import Diff_Build
+from osh.client.commands.cmd_build import Base_Build
 from osh.client.commands.common import (add_dist_git_url_option,
                                         add_tarball_option)
 
 
-class Mock_Build(Diff_Build):
+class Mock_Build(Base_Build):
     """analyze a SRPM"""
     enabled = True
     admin = False  # admin type account required
@@ -17,6 +17,15 @@ class Mock_Build(Diff_Build):
         # this option cannot be used for diff-build tasks
         add_tarball_option(self.parser)
         add_dist_git_url_option(self.parser)
+
+    def prepare_task_options(self, args, kwargs):
+        # Check if exactly one of SRPM, NVR, or dist-git URL is provided
+        if sum([bool(args), bool(kwargs.get("nvr")), bool(kwargs.get("git_url"))]) != 1:
+            self.parser.error("please specify either SRPM or NVR or dist-git URL")
+
+        self._process_srpm_option(args, kwargs)
+
+        return super().prepare_task_options(args, kwargs)
 
     def submit_task(self, options):
         return self.hub.scan.mock_build(options['mock_config'],
--- a/osh/client/commands/cmd_mock_build.py
+++ b/osh/client/commands/cmd_mock_build.py
@@ -2,7 +2,8 @@
 # SPDX-FileCopyrightText: Copyright contributors to the OpenScanHub project.
 
 from osh.client.commands.cmd_diff_build import Diff_Build
-from osh.client.commands.common import add_tarball_option
+from osh.client.commands.common import (add_dist_git_url_option,
+                                        add_tarball_option)
 
 
 class Mock_Build(Diff_Build):
@@ -15,6 +16,7 @@ class Mock_Build(Diff_Build):
 
         # this option cannot be used for diff-build tasks
         add_tarball_option(self.parser)
+        add_dist_git_url_option(self.parser)
 
     def submit_task(self, options):
         return self.hub.scan.mock_build(options['mock_config'],
--- a/osh/client/commands/common.py
+++ b/osh/client/commands/common.py
@@ -2,7 +2,9 @@
 # SPDX-FileCopyrightText: Copyright contributors to the OpenScanHub project.
 
 import json
-from optparse import OptionValueError
+import optparse
+
+from osh.common.validators import parse_dist_git_url
 
 
 def add_download_results_option(parser):
@@ -125,11 +127,27 @@ def add_nvr_option(parser):
     )
 
 
+def validate_git_url(option, opt_str, value, parser):
+
+    if value is None:
+        raise optparse.OptionValueError("Missing dist-git URL in request")
+    try:
+        # `ValueError` will be raised if an invalid dist-git URL is specified
+        parse_dist_git_url(value)
+    except ValueError as e:
+        raise optparse.OptionValueError(f"{e}")
+    else:
+        setattr(parser.values, option.dest, value)
+
+
 def add_dist_git_url_option(parser):
     parser.add_option(
         "--git-url",
         metavar="DIST_GIT_URL",
-        help="use a dist-git URL(specified by git-url) instead of a local file"
+        action="callback",
+        type="string",
+        callback=validate_git_url,
+        help="use a dist-git URL (specified by git-url) instead of a local file (EXPERIMENTAL)"
     )
 
 
@@ -168,7 +186,7 @@ def parse_json_option(option, opt, value, parser):
     try:
         json_value = json.loads(value)
     except json.JSONDecodeError:
-        raise OptionValueError(f"Option {opt}: invalid JSON value: {value}")
+        raise optparse.OptionValueError(f"Option {opt}: invalid JSON value: {value}")
 
     setattr(parser.values, option.dest, json_value)
 
--- a/osh/client/commands/common.py
+++ b/osh/client/commands/common.py
@@ -125,6 +125,14 @@ def add_nvr_option(parser):
     )
 
 
+def add_dist_git_url_option(parser):
+    parser.add_option(
+        "--git-url",
+        metavar="DIST_GIT_URL",
+        help="use a dist-git URL(specified by git-url) instead of a local file"
+    )
+
+
 def add_custom_model_option(parser):
     parser.add_option(
         "--cov-custom-model",
--- a/osh/client/completion/_osh-cli
+++ b/osh/client/completion/_osh-cli
@@ -145,7 +145,10 @@ _osh-cli_args()
             )
 
             if [[ "${words[1]}" == 'mock-build' ]]; then
-                args+=('--tarball-build-script=[sets command which should build the tarball]')
+                args+=(
+                    '--tarball-build-script=[sets command which should build the tarball]'
+                    '--git-url=[use a dist-git URL (specified by git-url) instead of a local file (EXPERIMENTAL)]'
+                )
             fi
             ;;
 
--- a/osh/hub/scan/scanner.py
+++ b/osh/hub/scan/scanner.py
@@ -485,10 +485,11 @@ class ClientScanScheduler(AbstractClientScanScheduler):
 
         # mock profile
         self.mock_config = get_or_fail('mock_config', self.options)
-        if self.mock_config == 'auto' and not self.build_nvr:
-            raise RuntimeError("'auto' mock config is only compatible with '--nvr'")
-        if self.mock_config == 'auto' and is_container_build(self.build_nvr, self.build_koji_profile):
-            self.mock_config = 'cspodman'
+        if self.mock_config == 'auto':
+            if not self.build_nvr and not self.dist_git_url:
+                raise RuntimeError("'auto' mock config is only compatible with '--nvr' or '--git-url'")
+            if self.dist_git_url is not None or is_container_build(self.build_nvr, self.build_koji_profile):
+                self.mock_config = 'cspodman'
         MockConfig.objects.verify_by_name(self.mock_config)
 
         self.comment = self.options.get('comment', '')
--- a/osh/hub/scan/scanner.py
+++ b/osh/hub/scan/scanner.py
@@ -17,6 +17,7 @@ from django.core.exceptions import ObjectDoesNotExist
 from kobo.django.upload.models import FileUpload
 from kobo.hub.models import TASK_STATES, Arch, Task
 
+from osh.common.validators import parse_dist_git_url
 from osh.hub.other.exceptions import PackageBlockedException
 from osh.hub.scan.check import (check_analyzers, check_build, check_nvr,
                                 check_obsolete_scan, check_package_is_blocked,
@@ -365,7 +366,7 @@ class AbstractClientScanScheduler:
 
     @classmethod
     def determine_priority(cls, entered_priority, supposed_nvr, srpm_name,
-                           is_tarball=False):
+                           dist_git_url=None, is_tarball=False):
         """determine priority of scheduled task
 
         :param entered_priority: priority submitted by client
@@ -392,10 +393,15 @@ class AbstractClientScanScheduler:
             if is_tarball:
                 srpm_name = re.sub(r'\.tar(\.[a-z0-9]+)?$', '', srpm_name)
 
-        try:
-            name_candidates.append(check_nvr(supposed_nvr or srpm_name)['name'])
-        except RuntimeError:
-            pass
+        if supposed_nvr or srpm_name:
+            try:
+                name_candidates.append(check_nvr(supposed_nvr or srpm_name)['name'])
+            except RuntimeError:
+                pass
+
+        if dist_git_url:
+            _, _, repo_name, _ = parse_dist_git_url(dist_git_url)
+            name_candidates.append(repo_name)
 
         if srpm_name:
             # try also only NV and the whole filename
@@ -422,9 +428,10 @@ class AbstractClientScanScheduler:
         if is_tarball:
             f = os.path.basename(filename)
             return f.rsplit(".", 2 if ".tar." in f else 1)[0]
-        # FIXME: resolve result_filename later
+
         if git_url is not None:
-            return None
+            _, _, repo_name, commit_hash = parse_dist_git_url(git_url)
+            return f"{repo_name}-{commit_hash}"
 
         raise RuntimeError("unknown input format of sources")
 
@@ -508,8 +515,6 @@ class ClientScanScheduler(AbstractClientScanScheduler):
         self.task_args['label'] = input_pkg
         self.task_args['method'] = self.method
         self.task_args['comment'] = self.comment
-        self.task_args['priority'] = AbstractClientScanScheduler.determine_priority(
-            self.priority, self.build_nvr, self.srpm_name, self.is_tarball)
         self.task_args['state'] = TASK_STATES['CREATED']
         self.task_args['args'] = {}
         if self.build_nvr:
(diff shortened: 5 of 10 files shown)


osh/client/commands/cmd_mock_build.py
```diff
@@ -15,6 +16,16 @@ def options(self):
 
         # this option cannot be used for diff-build tasks
         add_tarball_option(self.parser)
+        add_dist_git_url_option(self.parser)
+
+    def prepare_task_options(self, args, kwargs):
+        # Check if exactly one of SRPM, NVR, or dist-git URL is provided
+        if sum([bool(args), bool(kwargs.get("nvr")), bool(kwargs.get("git_url"))]) != 1:
```
hanchuntao, line 23: Is it possible to define two or more values?

osh/client/commands/cmd_mock_build.py
```diff
@@ -15,6 +16,16 @@ def options(self):
 
         # this option cannot be used for diff-build tasks
```
hanchuntao, line 17: Should we need to change the comment?

osh/client/commands/cmd_mock_build.py
```diff
@@ -15,6 +16,16 @@ def options(self):
...
         add_tarball_option(self.parser)
+        add_dist_git_url_option(self.parser)
+
+    def prepare_task_options(self, args, kwargs):
+        # Check if exactly one of SRPM, NVR, or dist-git URL is provided
+        if sum([bool(args), bool(kwargs.get("nvr")), bool(kwargs.get("git_url"))]) != 1:
+            self.parser.error("please specify either SRPM or NVR or dist-git URL")
+
+        self._process_srpm_option(args, kwargs)
```
hanchuntao, line 26: Can we add it to the `def options`, so that it looks more uniform

hanchuntao:
@rhyw  After installing the test rpm, run the command: `osh-cli mock-build --config cspodman --hub=https://covscan.lab.eng.brq2.redhat.com/osh/xmlrpc --git-url 'git+https://pkgs.devel.redhat.com/git/containers/amq-broker#d90c3cfb9c69e5fa023be1c08bd8ec98a876dfa3'`, I got the following error: https://privatebin.corp.redhat.com/?4f46af6856f23901#9sZE8pR8t8bwfBAuxUvLQLguCXCuMcUjRiMyxpixnwd1

kdudka:
@hanchuntao Did you also update the hub and worker?

Note that links to the RH internal pastebin instance are not much useful for external community members.

osh/client/commands/cmd_mock_build.py
rhyw: Prior to this change, the expected sole positional argument is either srpm or NVR. I don't think 2 or more values will be supported.

osh/client/commands/cmd_mock_build.py
rhyw: Not sure I understand. The changes in this MR doesn't change the behavior of `diff-build`, I think it's still true that the option cannot be used for diff-build.

osh/client/commands/cmd_mock_build.py
rhyw: It requires `args` and `kwargs` to be present for some validation, it may also populate the `kwargs`.

`options()` has no `args`/`kwarg` available, moving it there is probably not a good idea unless some careful refinement.

osh/client/commands/cmd_mock_build.py
rhyw: The expected behavior of being provided with more than 1 values should be that an error being raised.

osh/client/commands/cmd_mock_build.py
kdudka: I think `diff-build` would not work with `cspodman` anyway.  It should probably work with `version-diff-build` though.

osh/worker/csmock_runner.py
```diff
@@ -151,6 +166,13 @@ def srpm_download_analyze(self, analyzers, srpm_name, srpm_url, profile=None,
         srpm_path = self.download_file(srpm_url, srpm_name)
         return self.analyze(analyzers, srpm_path, profile, su_user, additional_arguments, **kwargs)
 
+    def dist_git_url_analyze(self, analyzers, dist_git_url, profile=None,
+                             su_user=None, additional_arguments=None, **kwargs):
+        if profile != "cspodman":
+            print("dist-git URL scan is incompatible with profile:", profile, file=sys.stderr)
+            return None, 2
+        return self.analyze(analyzers, dist_git_url, profile, su_user, additional_arguments, **kwargs)
```
kdudka, line 174: Where do we protect `dist_git_url` against shell command injection?  My quick check suggests that nowhere:
```sh
osh-cli mock-build --hub https://covscan.lab.eng.brq2.redhat.com/osh/xmlrpc --config=cspodman --git-url "https://example/ --help; ls -l ~/.config/configstore; #hash"
```

kdudka (changes requested):

hanchuntao:
> @hanchuntao Did you also update the hub and worker?
> 
> Note that links to the RH internal pastebin instance are not much useful for external community members.

yes, I update osh-client, osh-worker, but forget update osh-hub, after update it,  the command run correctly,and get correct result:
```
[root@chhan-rhel9 ~]# osh-cli mock-build --config cspodman --hub=https://covscan.lab.eng.brq2.redhat.com/osh/xmlrpc --git-url 'git+https://pkgs.devel.redhat.com/git/containers/amq-broker#d90c3cfb9c69e5fa023be1c08bd8ec98a876dfa3'
Task info: https://covscan.lab.eng.brq2.redhat.com/osh/task/467527/
Watching tasks (this may be safely interrupted)...
467527 MockBuild: FREE
--> FREE: 1 [total: 1]
467527 MockBuild: FREE -> OPEN (osh-test-worker-001.osh-001.preprod.iad2.dc.redhat.com)
--> OPEN: 1 [total: 1]
467527 MockBuild: OPEN (osh-test-worker-001.osh-001.preprod.iad2.dc.redhat.com) -> CLOSED (osh-test-worker-001.osh-001.preprod.iad2.dc.redhat.com)
--> CLOSED: 1 [total: 1]
[root@chhan-rhel9 ~]# 
```

osh/worker/csmock_runner.py
rhyw: There is a dist git URL check here: https://github.com/openscanhub/openscanhub/pull/278/files#diff-26bebd5671ea9c96d9eba3bdf8d8fcdcd3a4f8c594e359d29464c3623ff3aee3R36

But it appears the error was thrown before the validation is even triggerred. I'll see what can be done to throw this error earlier, probably to add some validation in cspodman cli.

kdudka: Validation in `osh-cli` is not a solution to the security issue because anyone can comment out the validation in their copy of `osh-cli` or just use the XML-RPC interface directly.

kdudka: ... and validation in `cspodman` would not help either because the shell injection happens in the shell that runs `cspodman`, not in the implementation of `cspodman`.

rhyw: I think it's still helpful from client side because we should always raise/return earlier. I had this commit addressed: https://github.com/openscanhub/openscanhub/pull/278/commits/69a902a156b4ab0b816a6bb15b81c62423094848

I'll see how similar sanitation/validation can be done from hub too. working in progress.

osh/client/commands/cmd_mock_build.py
rhyw: Latested updates applied both client/hub side validation.

kdudka (changes requested):
@rhyw Let's not over-complicate this.  The git URL is now checked at multiple places using multiple regular expressions.  Manually maintaining a set of disallowed characters is not a way to protect against shell command injection.  To protect against shell injection, please use `shlex.quote()` as we do for other arguments passed to `csmock` or `cspodman`.  To reject invalid URLs before starting a scan, I would just rewrite the regex that splits the URL parts so that it explicitly lists the set of allowed characters in each part of the URL.

rhyw:
@kdudka thanks for the suggestion, I have revised the way how validation was done a bit. The existing REs we use in csmock/cspodman are not enough to check the validity of the git URLs, some additional checks are added in latest code.

kdudka:
@rhyw Is this ready for review?

The commits seems to be in wrong order with some unnecessary changes back and forth:
- a41fb5340f07f52a5d06a947d1d8c668d3aa9612 introduces a call to the `parse_git_url()` function, which is not yet defined.
- The `parse_git_url()` function is defined later in commit ae3c67af9268b3a2b2cfd94eee20b2af4187e049.
- The `parse_git_url()` function is again removed in commit 0a4495883f09114e9cd7702125f29d39d8670447.

rhyw:
> @rhyw Is this ready for review?
> 
The pipeline is failed though I was checking the cause of the failure. 

> The commits seem to be in wrong order with some unnecessary changes back and forth:
> 
I had some problem rebasing earlier due to merge conflicts.

I'll mark this as Draft first, and mark it as Ready once these issues are fixed. Thanks

kdudka:
@rhyw Thanks for the update!  Unfortunately, the problem I pointed out in my last review does not seem to be resolved.  Commit d0b91ca30bc204f4d5aff7fbf0da4dc327470903 introduces a call to `parse_git_url()`, which is not defined.  Such inconsistencies make it difficult to use `git bisect` later on.

rhyw:
> @rhyw Thanks for the update! Unfortunately, the problem I pointed out in my last review does not seem to be resolved. Commit [d0b91ca](https://github.com/openscanhub/openscanhub/commit/d0b91ca30bc204f4d5aff7fbf0da4dc327470903) introduces a call to `parse_git_url()`, which is not defined. Such inconsistencies make it difficult to use `git bisect` later on.

This indeed was a problem, will fix

rhyw:
> _Commit [d0b91ca](https://github.com/openscanhub/openscanhub/commit/d0b91ca30bc204f4d5aff7fbf0da4dc327470903) introduces a call to parse_git_url(), which is not defined._

Fixed. I checked all 8 commits, I think all similar issues were addressed. Will make sure to keep the commits consistent in future PRs.

kdudka:
Regardless of our validation of the URL, which we do for splitting it into parts and providing more specific error messages, the URL needs to go through `shlex.quote()` before we append it to the shell command.  Only this way we can reliably protect OSH against shell injections through this field and for well-formed URLs the final sanitization should be harmless.

rhyw:
> the URL needs to go through shlex.quote()

Addressed in [1844997](https://github.com/openscanhub/openscanhub/pull/278/commits/184499781039313f8a95a5c1aa344e8660a10d91)

rhyw:
Added in 7b9075ed30a73481f14a6320fe1098d930cf52a9 the auto completion. Tested the completion in bash but not in zsh since I don't have it.

osh/worker/csmock_runner.py
```diff
@@ -171,7 +171,9 @@ def dist_git_url_analyze(self, analyzers, dist_git_url, profile=None,
         if profile != "cspodman":
             print("dist-git URL scan is incompatible with profile:", profile, file=sys.stderr)
             return None, 2
-        return self.analyze(analyzers, dist_git_url, profile, su_user, additional_arguments, **kwargs)
+        # sanitize the URL before appending to the shell command
+        sanitized_url = shlex.quote(dist_git_url)
+        return self.analyze(analyzers, sanitized_url, profile, su_user, additional_arguments, **kwargs)
```
hanchuntao, line 176: We can add `shlex.quote` in line 176 directly.

osh/worker/csmock_runner.py
```diff
@@ -100,19 +100,18 @@ def do(self, command, output_path=None, su_user=None, **kwargs):
...
-                result_filename=None, profile_url=None, **kwargs):
+    def determine_output_path(self, srpm_path, result_filename):
         if result_filename is None:
             result_filename = os.path.basename(srpm_path)[:-8]
-        if self.tmpdir:
-            output_path = os.path.join(self.tmpdir, result_filename + '.tar.xz')
-        else:
-            output_path = os.path.join(os.getcwd(), result_filename + '.tar.xz')
+        output_path = os.path.join(self.tmpdir if self.tmpdir else os.getcwd(),
+                                   result_filename + '.tar.xz')
```
hanchuntao, line 107: The `result_filename` maybe `None`, so we cannot use `None` + '.tar.xz'

osh/worker/csmock_runner.py
```diff
@@ -130,16 +129,32 @@ def analyze(self, analyzers, srpm_path, profile=None, su_user=None, additional_a
         if output_path:
             cmd += ' -o ' + shlex.quote(output_path)
 
+        return cmd
+
+    def handle_additional_arguments(self, additional_arguments):
```
hanchuntao, line 134: This is a static method

osh/client/commands/cmd_build.py
```diff
@@ -81,6 +82,8 @@ def check_build(self, args, kwargs, prefix=""):
             if result is not None:
                 self.parser.error(result)
             options[prefix + "brew_build"] = nvr
+        elif dist_git_url is not None:
```
hanchuntao, line 85: `''` also is not `None`

osh/worker/csmock_runner.py
rhyw: Yeah that works, but I'd like to preserve the comment line L174, it's probably helpful.

osh/worker/csmock_runner.py
rhyw: I'm not sure I understand what you mean?

osh/client/commands/cmd_build.py
rhyw: No I think `dist_git_url` won't be an empty string here

osh/worker/csmock_runner.py
rhyw: it won't because L105 makes sure it's not. But this indeed remind me of an issue, i.e., 
```
    def determine_output_path(self, srpm_path, result_filename):
        if result_filename is None:
            result_filename = os.path.basename(srpm_path)[:-8]
        output_path = os.path.join(self.tmpdir if self.tmpdir else os.getcwd(),
                                   result_filename + '.tar.xz')
```
I guess the slicing(`[:8]`) is to remove the trailing `.src.rpm` in the filename. In case of dist-git url, The commit hash would be truncated. I'll improve this.

kdudka:
@rhyw Could you please rebase this to fix the CI failures?

osh/worker/csmock_runner.py
rhyw: I had a second check, it won't be truncated since `result_filename` is processed elsewhere:
https://github.com/openscanhub/openscanhub/pull/278/files#diff-2ae41bcd4e6cf6150fb0c793f090773acb92488f9922cfae2529518523ab43e0R432-R434

rhyw:
@kdudka rebased, the ci is green now. Have another test on stage and appears to be working as expected.

osh/worker/csmock_runner.py
```diff
@@ -151,6 +166,15 @@ def srpm_download_analyze(self, analyzers, srpm_name, srpm_url, profile=None,
         srpm_path = self.download_file(srpm_url, srpm_name)
         return self.analyze(analyzers, srpm_path, profile, su_user, additional_arguments, **kwargs)
 
+    def dist_git_url_analyze(self, analyzers, dist_git_url, profile=None,
+                             su_user=None, additional_arguments=None, **kwargs):
+        if profile != "cspodman":
+            print("dist-git URL scan is incompatible with profile:", profile, file=sys.stderr)
+            return None, 2
+        # sanitize the URL before appending to the shell command
+        sanitized_url = shlex.quote(dist_git_url)
```
kdudka, line 175: Could you please squash the commit f0e56d5093a051e50cd55b6b85af326d786bd5d8, which introduces the sanitization?  Since this was discovered in the review, I do not want a commit that fixes a shell injection issue introduced by a previous commit in the final branch.

kdudka (changes requested):
@rhyw Thanks for the update!  A few suggestions inline...

osh/worker/csmock_runner.py
rhyw: done

osh/worker/csmock_runner.py
```diff
@@ -151,6 +166,15 @@ def srpm_download_analyze(self, analyzers, srpm_name, srpm_url, profile=None,
         srpm_path = self.download_file(srpm_url, srpm_name)
         return self.analyze(analyzers, srpm_path, profile, su_user, additional_arguments, **kwargs)
 
+    def dist_git_url_analyze(self, analyzers, dist_git_url, profile=None,
+                             su_user=None, additional_arguments=None, **kwargs):
+        if profile != "cspodman":
```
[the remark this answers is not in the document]
rhyw: it won't:
```
  ...
  File "/usr/lib/python3.6/site-packages/osh/hub/scan/scanner.py", line 452, in __init__
    self.validate_options()
  File "/usr/lib/python3.6/site-packages/osh/hub/scan/scanner.py", line 489, in validate_options
    raise RuntimeError("'auto' mock config is only compatible with '--nvr'")
RuntimeError: 'auto' mock config is only compatible with '--nvr'
```

Is this the intended behavior?

osh/hub/scan/scanner.py
```diff
@@ -452,16 +462,21 @@ def validate_options(self):
...
-            self.srpm_path = check_srpm_response['srpm_path']
-            self.srpm_name = check_srpm_response['srpm_name']
+        if any((self.build_nvr, self.upload_id)):
+            check_srpm_response = check_srpm(self.upload_id, self.build_nvr, self.username, self.is_tarball)
+            if check_srpm_response['type'] == 'build':
+                self.build_koji_profile = check_srpm_response['koji_profile']
+            elif check_srpm_response['type'] == 'upload':
+                self.srpm_path = check_srpm_response['srpm_path']
+                self.srpm_name = check_srpm_response['srpm_name']
+        else:
```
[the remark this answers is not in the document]
rhyw: resolved

osh/client/completion/_osh-cli
```diff
@@ -145,7 +145,10 @@ _osh-cli_args()
             )
 
             if [[ "${words[1]}" == 'mock-build' ]]; then
-                args+=('--tarball-build-script=[sets command which should build the tarball]')
+                args+=(
+                    '--tarball-build-script=[sets command which should build the tarball]'
+                    '--git-url=[use a dist-git URL(specified by git-url)instead of a local file]'
```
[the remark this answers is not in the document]
rhyw: updated.

osh/client/commands/common.py
```diff
@@ -125,6 +127,30 @@ def add_nvr_option(parser):
...
+
+
+def add_dist_git_url_option(parser):
+    parser.add_option(
+        "--git-url",
+        metavar="DIST_GIT_URL",
+        action="callback",
+        type="string",
+        callback=validate_git_url,
+        help="use a dist-git URL(specified by git-url) instead of a local file"
```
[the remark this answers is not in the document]
rhyw: done. didn't know EXPERIMENTAL could of this meaning.. thx

osh/worker/csmock_runner.py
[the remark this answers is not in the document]
kdudka: It would be better to make `osh-cli` accept `--config=auto` with `--git-url`, too.  Given the fact, that `--git-url` works for container builds only, we can always translate `auto` to `cspodman` like we do for Brew builds:
- https://github.com/openscanhub/openscanhub/blob/ee9fd3ed9fbb75b2825b62bbf5edc103996aebc3/osh/hub/scan/scanner.py#L156
- https://github.com/openscanhub/openscanhub/blob/ee9fd3ed9fbb75b2825b62bbf5edc103996aebc3/osh/hub/scan/scanner.py#L246

The long term plan is to make `osh-cli` use `--config=auto` (instead of `fedora-rawhide-x86_64`) by default so that users do not need to use the option in 99% of cases.

osh/client/commands/common.py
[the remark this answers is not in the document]
kdudka: I read `(EXPERIMENTAL)` as `may not work for everybody`.

osh/worker/csmock_runner.py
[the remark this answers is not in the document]
rhyw: I've addressed it in https://github.com/openscanhub/openscanhub/pull/278/commits/0c8d6e52a06a970d21ff4a072dde079714797791 . had some tests on stage, works fine with `--config=auto`.

However this change only impacts `ClientScanScheduler`. The original jira issue doens't mention diff-build or tasks from errata, thus the PR doesn't contain support for `git-url` with diff-build. I had a quick check, to support that, there'll a bit more clumsy logic needed.

Let me know if it's needed.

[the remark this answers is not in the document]
kdudka: Errata Tool scans Brew builds only.  The support for `version-diff-build` tasks can be introduced later if needed.

osh/hub/scan/scanner.py
```diff
@@ -471,10 +486,14 @@ def validate_options(self):
...
-        if self.mock_config == 'auto' and is_container_build(self.build_nvr, self.build_koji_profile):
-            self.mock_config = 'cspodman'
+        if self.mock_config == 'auto':
+            if not self.build_nvr and not self.dist_git_url:
+                raise RuntimeError("'auto' mock config is only compatible with '--nvr' or '--dist-git-url'")
+            is_container = (self.build_nvr is not None
+                            and self.build_koji_profile is not None
+                            and is_container_build(self.build_nvr, self.build_koji_profile))
+            if is_container or self.dist_git_url is not None:
+                self.mock_config = 'cspodman'
```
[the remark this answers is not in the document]
rhyw: With `--git-url` option, the `self.build_nvr`/`self.build_koji_profile` could be None thus causes the `is_container_build` to throw an exception.

I could simplify it this way and update the `is_container_build` to handle the `None` param there.

osh/hub/scan/scanner.py
```diff
@@ -471,10 +486,14 @@ def validate_options(self):
 
         # mock profile
         self.mock_config = get_or_fail('mock_config', self.options)
-        if self.mock_config == 'auto' and not self.build_nvr:
-            raise RuntimeError("'auto' mock config is only compatible with '--nvr'")
-        if self.mock_config == 'auto' and is_container_build(self.build_nvr, self.build_koji_profile):
-            self.mock_config = 'cspodman'
+        if self.mock_config == 'auto':
+            if not self.build_nvr and not self.dist_git_url:
+                raise RuntimeError("'auto' mock config is only compatible with '--nvr' or '--dist-git-url'")
```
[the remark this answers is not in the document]
rhyw: My BAD! Thanks for spotting

osh/hub/scan/scanner.py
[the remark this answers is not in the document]
kdudka: If the `--git-url` option is used, `self.dist_git_url is not None` evaluates to `True`.  Consequently, the second operand of the `or` operator is not evaluated and `is_container_build()` is not called (and thus cannot throw an exception).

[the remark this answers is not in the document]
rhyw: Exactly! Updated, thanks

kdudka (approved):
@rhyw Thanks for the update!  I think we can merge this...

rhyw:
Thanks for all the review!

PR #280, opened 2024-05-29
hub: handle `rhel_version` option for ASYNC ET advisories

lzaoral:
Resolves: https://issues.redhat.com/browse/OSH-69

### DIFF
--- a/docs/scanning_releases.md
+++ b/docs/scanning_releases.md
@@ -17,9 +17,9 @@ Here's a scan request from Errata Tool:
 }
 ```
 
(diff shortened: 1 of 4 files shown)


hanchuntao:
Some checks were not successful

lzaoral:
@hanchuntao That was an unrelated issue in Testing Farm.  Re-run fixed it.

kdudka (approved):
The code changes look good.  We yet need to think how to configure this so that it works as expected for all the supported products.  For example when release is `RHOSE ASYNC - AUTO` and `rhel_version` is `RHEL-8` or `RHEL-9`.  We want to handle such requests in a way that we do not compare apples with oranges.

PR #283, opened 2024-06-19
hub: fix generation of mock configs for `noarch` builds

lzaoral:


osh/hub/scan/mock.py
```diff
@@ -173,6 +187,9 @@ def _create_mock_config(tag, arch, koji_profile, dest_dir):
     extra_repos = '\n'.join(matched_repos).replace('\n', '\\n')
     contents = contents.replace("[main]", extra_repos + '\\n\\n[main]')
 
+    # use $basearch for the "[build]" repo
+    contents = re.sub(rf'({tag}/latest/){arch}', r'\1$basearch', contents)
```
kdudka, line 191: Should we use the `rf` vs. `fr` string prefix consistently across the surrounding commands?

kdudka (approved):

osh/hub/scan/mock.py
lzaoral: Addressed in ea6846623d9856043b9b7edf37cb1a2b885c6c18.

PR #285, opened 2024-06-26
allow user to specify task metadata from client

rhyw:
Related: https://issues.redhat.com/browse/OSH-567
Closes: https://github.com/openscanhub/openscanhub/pull/285

### DIFF
--- a/osh/client/commands/common.py
+++ b/osh/client/commands/common.py
@@ -1,6 +1,10 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 # SPDX-FileCopyrightText: Copyright contributors to the OpenScanHub project.
 
+import json
+from optparse import OptionValueError
+
+
 def add_download_results_option(parser):
     parser.add_option(
         "-d",
@@ -149,4 +153,24 @@ def add_tarball_option(parser):
              "then the tarball will be scanned. "
              "This option sets command which should build the package, usually this should be just "
              "\"make\", in case of packages which doesn't need to be built, just pass \"true\".",
+    )
+
+
+def parse_json_option(option, opt, value, parser):
+    try:
+        json_value = json.loads(value)
+    except json.JSONDecodeError:
+        raise OptionValueError(f"Option {opt}: invalid JSON value: {value}")
+
+    setattr(parser.values, option.dest, json_value)
+
+
+def add_task_metadata_option(parser):
+    parser.add_option(
+        "--metadata",
+        dest="metadata",
+        action="callback",
+        callback=parse_json_option,
+        help="Specify task metadata as a JSON string, e.g., "
+             "--metadata='{\"product\": \"RHEL\", \"version\": \"9.3\"}'"
     )
--- a/osh/client/completion/_osh-cli
+++ b/osh/client/completion/_osh-cli
@@ -125,6 +125,7 @@ _osh-cli_args()
                 '--config=[specify mock config name]:osh:->configs'
                 '--comment=[a task description]'
                 '--json[print created task info in JSON]'
+                '--metadata=[specify task metadata as a JSON string]'
                 '--task-id-file=[task id is written to this file]:filename:_files'
                 "--nowait[don't wait until tasks finish]"
                 # '*' enables repetition
--- a/osh/hub/scan/check.py
+++ b/osh/hub/scan/check.py
@@ -5,6 +5,7 @@
 Functions related to checking provided data
 """
 
+import json
 import logging
 import os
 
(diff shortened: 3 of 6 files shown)


rhyw:
So the original request asks for the option for `mock-build`, I think it probably makes more sense to make it available for  `mock-build`, `diff-build`, `version-diff-build`, thus the patch.

We don't have filtering capabilities for task args for `find-tasks` subcommand. I had some local test, it supports options like `args__contains`, to filter values for a specific key in `args` is not supported right now.

kdudka (approved):
I have not tested it myself but at first glance it looks sane.

rhyw:
added one more commit such that `--metadata` is available in command completion.

osh/client/commands/common.py
```diff
@@ -150,3 +154,24 @@ def add_tarball_option(parser):
              "This option sets command which should build the package, usually this should be just "
              "\"make\", in case of packages which doesn't need to be built, just pass \"true\".",
     )
+
+
+def parse_json_option(option, opt, value, parser):
+    try:
+        json_value = json.loads(value)
+    except json.JSONDecodeError as e:
+        raise OptionValueError(f"Option {opt}: invalid JSON value: {e}")
```
lzaoral, line 163: This is a good idea but the error message looks a bit too convoluted:
```console
$ osh-cli mock-build --metadata=
Usage: osh-cli mock-build [options] <args>

osh-cli: error: Option --metadata: invalid JSON value: Expecting value: line 1 column 1 (char 0)
```

osh/hub/scan/check.py
```diff
@@ -121,3 +122,15 @@ def check_srpm(upload_id, build_nvr, task_user, is_tarball=False):
...
+
+def check_task_metadata(metadata):
+    if isinstance(metadata, dict):
+        return metadata
+    elif isinstance(metadata, str):
+        try:
+            return json.loads(metadata)
+        except json.JSONDecodeError as e:
+            raise RuntimeError(f"Invalid JSON value: {e}")
+    else:
```
lzaoral, line 135: This `else` is redundant because all previous blocks end with a jump statement.

osh/hub/scan/check.py
```diff
@@ -121,3 +122,15 @@ def check_srpm(upload_id, build_nvr, task_user, is_tarball=False):
         }
 
     raise RuntimeError('No source RPM or tarball specified.')
+
+
+def check_task_metadata(metadata):
+    if isinstance(metadata, dict):
+        return metadata
+    elif isinstance(metadata, str):
```
lzaoral, line 130: This can be a plain `if` because all previous blocks end with a jump statement.

osh/hub/scan/check.py
```diff
@@ -121,3 +122,15 @@ def check_srpm(upload_id, build_nvr, task_user, is_tarball=False):
...
+        try:
+            return json.loads(metadata)
+        except json.JSONDecodeError as e:
+            raise RuntimeError(f"Invalid JSON value: {e}")
```
lzaoral, line 134: You can simplify this to `RuntimeError("Invalid JSON value:", e)`

osh/hub/scan/scanner.py
```diff
@@ -487,6 +488,8 @@ def validate_options(self):
         self.client_csmock_args = self.options.get('csmock_args', None)
 
         self.metadata = self.options.get('metadata')
+        if self.metadata:
+            self.metadata = check_task_metadata(self.metadata)
```
lzaoral: nit: You can move this check to `check_task_metadata` as well.

lzaoral (changes requested):
Thank you, @rhyw!  I have tested the feature and it seems to work as expected!  Some comments inline.

osh/client/commands/common.py
rhyw: I've updated the exception to print the original json string instead. Thanks for suggestion!

osh/client/commands/common.py
```diff
@@ -150,3 +154,24 @@ def add_tarball_option(parser):
...
+
+    setattr(parser.values, option.dest, json_value)
+
+
+def add_task_metadata_option(parser):
+    parser.add_option(
+        "--metadata",
+        dest="metadata",
+        type="string",
```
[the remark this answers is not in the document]
rhyw: This line's removed. I think it's ok to keep it too.

osh/client/completion/_osh-cli
```diff
@@ -125,6 +125,7 @@ _osh-cli_args()
                 '--config=[specify mock config name]:osh:->configs'
                 '--comment=[a task description]'
                 '--json[print created task info in JSON]'
+                '--metadata[specify task metadata as a JSON string]'
```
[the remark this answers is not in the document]
rhyw: Good catch, this indeed is an issue, fixed.

osh/hub/scan/scanner.py
rhyw: This is reasonable in some way. I'd like to keep it as it as so the `check_task_metadata` only deals with non-null values.

rhyw:
@lzaoral Thanks for testing. Patch updated per suggestions.

lzaoral (approved):

rhyw:
Thanks for the review!

PR #287, opened 2024-06-28
hub: fix `kpatch-patch` not being blocked with `auto` mock configs

lzaoral:
Related: https://issues.redhat.com/browse/OSH-69


kdudka (approved):
Looks good.  Thanks!

PR #289, opened 2024-07-04
packaging: use `python3dist` to specify django dependency

lzaoral:
... to make OSH compatible with versioned Django packages in EPEL and Fedora.


kdudka (approved):
Looks good.  Thanks!

PR #292, opened 2024-08-02
hub: fix a sentence in the homepage

siteshwar:
Some of these plugins may be disabled, so this sentence is not always accurate.

### DIFF
--- a/osh/hub/templates/index-content.html
+++ b/osh/hub/templates/index-content.html
@@ -5,7 +5,7 @@ OpenScanHub is a service that runs various static analyzers on RPM packages. The
 </p>
 
 <p>
-OpenScanHub by default uses <a href="https://github.com/danmar/cppcheck">Cppcheck</a>, <a
+OpenScanHub by default can use <a href="https://github.com/danmar/cppcheck">Cppcheck</a>, <a
     href="https://www.shellcheck.net/">ShellCheck</a>, the static analyzers embedded in <a
     href="https://developers.redhat.com/blog/2020/03/26/static-analysis-in-gcc-10">GCC</a> and <a
     href="https://clang-analyzer.llvm.org/">Clang</a>, and the <a


osh/hub/templates/index-content.html
```diff
@@ -5,7 +5,7 @@ <h2>OpenScanHub Web User Interface</h2>
 </p>
 
 <p>
-OpenScanHub by default uses <a href="https://github.com/danmar/cppcheck">Cppcheck</a>, <a
+OpenScanHub by default can use <a href="https://github.com/danmar/cppcheck">Cppcheck</a>, <a
```
kdudka, line 8: I do not think the proposed change improves anything.  The original sentence says: `When you submit a scan, OSH will use these scanners unless you configure the set of enabled scanners manually.`

Are you trying to highlight that Cppcheck will not be used for Python code etc.?

siteshwar: This was a follow up on https://github.com/packit/packit.dev/pull/914#discussion_r1698427411. We do not enable `find-unicode-control` by default in the [fixtures](https://github.com/openscanhub/openscanhub/blob/311382e5a9df1065d198f4c9592ae86f07bababe/osh/hub/scan/fixtures/initial_data.json#L476).

kdudka: I think that `find-unicode-control` can be added to the fixtures but it will not affect the already running OSH instances (unlike the proposed change of the template).

siteshwar: Shall I also add `gitleaks` tool in the fixtures?

kdudka: I am not sure.  `gitleaks` is currently not enabled by default on any of the maintained OSH instances.

siteshwar: I would close this as it seems there is no clear idea about which words to use on the homepage, and which analyzers to enable by default in the fixtures.

PR #293, opened 2024-08-07
Fixes for CI

siteshwar:


containers/scripts/deploy.sh
```diff
@@ -13,9 +13,6 @@ CONTAINERS=(
     osh-worker
 )
 
-if [ -z "$IS_PODMAN" ]; then
-    PROFILE='--profile=full-dev'
-fi
```
kdudka, line 18: Does this really change anything in the CI environment?  I would expect the condition to evaluate to `false` in CI.  I am afraid this could break local development environment for MacOS users.

kdudka: Because macOS is the environment where the condition evaluates to `true` if I remember correctly.

siteshwar: This variable was meant to be used for docker. I have removed the profile as well as the use of this variable from rest of the code.

@lzaoral Could you verify this does not break anything on macOS?

lzaoral: @kdudka: @siteshwar is correct here, the condition evaluates to true on any platform that uses `docker` because `--profile` feature of `docker-compose` was not supported in `podman-compose` at the time this script was written.  Note that I've never tried `docker` on macOS due to licensing issues (and `podman` is my tool of choice anyway) so the condition would never evaluate to true on my machine.

lzaoral:
@siteshwar If you want to remove profiles from `compose.yaml`, you should remove the code depends on it as well: https://github.com/search?q=repo%3Aopenscanhub%2Fopenscanhub%20full-dev&type=code

containers/scripts/deploy.sh
kdudka: @jperezdealgaba Could you please check whether the changes from this pull request do no break the development environment on macOS with docker (assuming you are still using it)?

siteshwar:
> @siteshwar If you want to remove profiles from `compose.yaml`, you should remove the code depends on it as well: https://github.com/search?q=repo%3Aopenscanhub%2Fopenscanhub%20full-dev&type=code

The `--full-dev` option adds `osh-client` container to the list of containers to be run. It can be used even if we remove the profile.

lzaoral:
I did not know that.  In that case the profile spec in `compose.yaml` was redundant from the start...

containers/scripts/deploy.sh
jperezdealgaba: [osh.log](https://github.com/user-attachments/files/16539350/osh.log)
I have tested this branch and it seems to work as expected.
I have also attacked the logs in case they are needed

kdudka: @jperezdealgaba Thank you for testing it!

kdudka (approved):

PR #294, opened 2024-08-08
hub: fix arch detection of mock configs for `wrapperRPM` tasks

lzaoral:
Fixes: 8bb2b03 ("hub: fix generation of mock configs for noarch builds)
Related: https://issues.redhat.com/browse/OSH-69
Related: https://issues.redhat.com/browse/OSH-538

### DIFF
--- a/osh/hub/scan/mock.py
+++ b/osh/hub/scan/mock.py
@@ -239,6 +239,11 @@ def generate_mock_configs(nvr, koji_profile):
     if not koji_proxy.getTag(tag):
         raise RuntimeError(f'Tag {tag} does not exists in {koji_profile}!')
 
+    # get arches of given tag
+    arches = _get_build_arches(koji_proxy, task['id'], tag)
+    if not arches:
+        raise RuntimeError(f'No arches found for tag "{tag}"!')
+
     # generate a config for every built arch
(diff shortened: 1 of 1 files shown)


osh/hub/scan/mock.py
```diff
@@ -116,23 +116,30 @@ def _get_build_arches(koji_proxy, task_id, tag):
...
         child = koji_proxy.getTaskInfo(child_id, request=True)
 
         # skip non-build subtasks
         method = child['method']
-        if method != 'buildArch':
+        if method not in ('buildArch', 'wrapperRPM'):
             continue
 
+        # wrapperRPM tasks do not track arches directly
+        if method == 'wrapperRPM':
```
kdudka, line 134: If this was inserted before the previous condition, the other condition could stay as it is.

kdudka (approved):
Looks good.  Thanks for the quick fix!

osh/hub/scan/mock.py
lzaoral: Good idea!

lzaoral: Addressed in 2fe56b90291e32119ef69f8689b428c7424d1f44.

PR #296, opened 2024-08-09
client: explicitly specify expected --metadata type

rhyw:
otherwise the callback function is unable to correctly handle the value passed to the option, raising errors like this:

osh-cli: error: --metadata option does not take a value

Related: https://issues.redhat.com/browse/OSH-567


rhyw:
Fix an issue from https://github.com/openscanhub/openscanhub/pull/285

kdudka (approved):
Looks good.  Thanks!

Issue #297, opened 2024-08-14
Use the term "findings" instead of "defects" in the reports


siteshwar:
This is a follow up on https://github.com/csutils/csdiff/issues/193.

There are places in the OSH code that use the term "defects" instead of "findings" for user visible reports. This issue tracks renaming this term in user visible reports.

PR #298, opened 2024-09-03
hub: use the term "findings"

siteshwar:
... instead of "defects" in the reports.

Resolves: https://github.com/openscanhub/openscanhub/issues/297


kdudka (approved):
Assuming the waiving interface and statistics are out of scope, I think this looks good.

PR #301, opened 2024-10-17
hub: update message topics for fedora

siteshwar:
... and use only `started` and `finished` topics for each task.

Related: https://github.com/openscanhub/fedora-infra/issues/3

### DIFF
--- a/osh/hub/osh_xmlrpc/worker.py
+++ b/osh/hub/osh_xmlrpc/worker.py
@@ -133,7 +133,7 @@ def finish_task(request, task_id):
         scan_results_js_path += "?format=raw"
         scan_results_js_url = request.build_absolute_uri(scan_results_js_path)
 
-        body = {'scan-results.js': scan_results_js_url}
+        body = {'status': 'success', 'scan-results.js': scan_results_js_url}
         if base_task:
             added_js_path = django.urls.reverse("task/log", args=[task_id, "added.js"])
             added_js_path += "?format=raw"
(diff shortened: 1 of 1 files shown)


siteshwar:
@lbarcziova @majamassarini Review?

majamassarini (approved):
Thanks a lot!

kdudka (approved):
Looks good to me.  Thanks!  Just to make sure, are we using verbs instead of adjectives to denote the states on purpose?

siteshwar:
> Looks good to me. Thanks! Just to make sure, are we using verbs instead of adjectives to denote the states on purpose?

I have just put the task state from kobo in the `status` field. Let me know if they should be written differently.

kdudka:
The [kobo states](https://github.com/release-engineering/kobo/blob/ea882524c843ece43d2193f79eef914c094ba393/kobo/client/constants.py#L14) are named `CANCELED`, `FAILED`, and `INTERRUPTED` whereas this pull request uses `cancel`, `fail`, and `interrupt`.  It is up to you.  I am fine with both the variants...

siteshwar:
> The [kobo states](https://github.com/release-engineering/kobo/blob/ea882524c843ece43d2193f79eef914c094ba393/kobo/client/constants.py#L14) are named `CANCELED`, `FAILED`, and `INTERRUPTED` whereas this pull request uses `cancel`, `fail`, and `interrupt`. It is up to you. I am fine with both the variants...

I would leave them as it is. Thanks!

PR #311, opened 2025-01-23
packaging: install daemons to `/usr/bin`

kdudka:
`/usr/sbin` is a symlink to `/usr/bin` on Fedora 42 and there is little to no benefit in using `/usr/sbin` on older Fedora/RHEL releases anyway.

Temporarily provide the `/usr/sbin/osh-worker` symlink on systems with separate `/usr/sbin` to support reexec upon upgrade of the `osh-worker` RPM package.

Related: https://fedoraproject.org/wiki/Changes/Unify_bin_and_sbin
Resolves: https://bugzilla.redhat.com/2340988


kdudka:
Actually, I will keep the `/usr/sbin/osh-worker` path in `osh-worker.service` in order to keep the reexec on reload feature work: https://github.com/release-engineering/kobo/blob/b4b486f53bdfa091bbe2a6a1049b204b942b345b/kobo/worker/main.py#L97

kdudka:
@siteshwar I am also thinking about releasing `osh-1.1.0` when this gets merged.  The last OSH release is from April 2024.

siteshwar (approved):

kdudka:
@siteshwar Thanks for review!

PR #313, opened 2025-01-28
packaging: release osh-1.1.1

kdudka:
While releasing `osh-1.1.0` last week I forgot to update the version in `setup.py` and the change log in the upstream SPEC file.  This minor release fixes it.


kdudka:
I have added commit b8d2302259e52e430cfe5f4a1ad4b4385227abb9 to improve the CI.  If it does not fully fix the issue, it will make it at least easier to debug when it happens in the future.

siteshwar (approved):

kdudka:
@siteshwar Thanks for review!

PR #295, opened 2024-08-08
codeowners: remove @lzaoral

lzaoral:
The RHEL-related responsibilities take the majority of my work-time and I don't have time nor the motivation to be the main reviewer for this project.

From now on, I'd like to officially limit my contributions only to critical regressions which is the actual state of things for few past months anyway.  I'll try to finish #243 soon but other QOL improvements or (much needed) refactoring is out of scope for me.

PR #318, opened 2025-06-18
Update OSH URLs to new hostname

sfowl:
Not sure if this change is strictly necessary but a new commit on main will be helpful for propagating new internal builds of `osh-client`, which contain the `/etc/osh/client.conf` file with the updated `HUB_URL`.


osh/client/osh-cli
```diff
@@ -36,9 +36,11 @@ def handle_deprecated_options(conf):
               file=sys.stderr)
 
     # xmlrpclib/httplib do not follow HTTP redirect
-    if "cov01.lab.eng.brq.redhat.com" in conf["HUB_URL"]:
-        conf["HUB_URL"] = "https://cov01.lab.eng.brq2.redhat.com/osh/xmlrpc"
-        print_warning("HUB_URL")
+    hosts = ("cov01.lab.eng.brq.redhat.com", "cov01.lab.eng.brq2.redhat.com", "osh.prodsec.redhat.com")
```
kdudka, line 39: The reason for this hard-coded transformation is that `client.conf` is packaged as `%config(noreplace)`: https://github.com/openscanhub/openscanhub/blob/55f0214d2656a1e7a74dd65c18520dbe1d350a90/osh.spec#L221

This means that users who changed the `client.conf` file locally will not start using the updated `HUB_URL` after installing the updated RPM package until they merge the configuration file changes manually.  Therefore, I would not put `osh.prodsec.redhat.com` into this tuple.  It is unlikely that users would set such a `HUB_URL` manually in their configuration file and, if they do, they should use a correct URL with no need to rewrite it.

kdudka (changes requested):
Please use the `client: ` prefix in the commit message summary as described in [CONTRIBUTING.md](https://github.com/openscanhub/openscanhub/blob/main/CONTRIBUTING.md).

osh/client/osh-cli
sfowl: Makes sense, removed from tuple

kdudka (approved):
Looks good.  Thanks for the update!  Now we need to figure out why the upstream CI started to fail...

sfowl:
> Looks good. Thanks for the update! Now we need to figure out why the upstream CI started to fail...

I think the cause was some change in behaviour between podman-compose 1.3.0 (used in last successful [PR](https://github.com/openscanhub/openscanhub/actions/runs/14081702409/job/39681962464)) and [1.4.0](https://github.com/containers/podman-compose/releases/tag/v1.4.0).

containers/scripts/deploy.sh
```diff
@@ -44,7 +44,7 @@ main() {
     prepare_deploy
 
     # build the containers but do not start them
-    podman-compose -p osh up --build --no-start "${CONTAINERS[@]}"
+    podman-compose -p osh build "${CONTAINERS[@]}"
```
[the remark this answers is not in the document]
sfowl: It works for me locally both with versions 1.2.0 and 1.4.0. The `build` cmd at least seems to do the same thing in 1.0.0 as well, but I can't easily test an entire integration test run with that version.

FWIW though, I've encountered similar issues with podman-compose in the past (changes in behaviour / breakages).

siteshwar (approved):

kdudka:
@sfowl Thank you for fixing the CI!  I have only swapped the commits such that the CI fix goes first (to make the history more `git bisect`-friendly).

lzaoral:
> [!WARNING]  
> The following changes are mostly untested and I have no intention to submit them to `kobo` myself because I want to keep my involvement with this project at minimum (see #295).

@kdudka brought the F42 TF failures to my attention.  They are caused by the following change in Django 5.2: https://github.com/django/django/commit/50f89ae850f6b4e35819fe725a08c7e579bfd099

The following `kobo` patch removes the usage of `MiddlewareMixin` (which was mainly needed for Django 1.9 and older but `kobo` no longer supports these releases) and replaces them with a bit more modern alternative:

```patch
diff --git a/kobo/django/auth/middleware.py b/kobo/django/auth/middleware.py
index 864d46d..212f2b9 100644
--- a/kobo/django/auth/middleware.py
+++ b/kobo/django/auth/middleware.py
@@ -1,9 +1,8 @@
 from django.contrib.auth.middleware import RemoteUserMiddleware
-from django.utils.deprecation import MiddlewareMixin

 from kobo.django.helpers import call_if_callable

-class LimitedRemoteUserMiddleware(RemoteUserMiddleware, MiddlewareMixin):
+class LimitedRemoteUserMiddleware(RemoteUserMiddleware):
     '''
     Same behaviour as RemoteUserMiddleware except that it doesn't logout user
     if is already logged in.
diff --git a/kobo/django/menu/middleware.py b/kobo/django/menu/middleware.py
index e118edc..ad9438d 100644
--- a/kobo/django/menu/middleware.py
+++ b/kobo/django/menu/middleware.py
@@ -1,7 +1,5 @@
 # -*- coding: utf-8 -*-

-from django.utils.deprecation import MiddlewareMixin
-
 from kobo.django.menu import menu


@@ -20,10 +18,18 @@ class LazyMenu(object):
         return request._cached_menu


-class MenuMiddleware(MiddlewareMixin):
+class MenuMiddleware():
     """
     @summary: Middleware for menu object.
     """
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        self.process_request(request)
+        response = self.get_response(request)
+        return response
+
     def process_request(self, request):
         """
         @summary: Adds menu to request object
diff --git a/kobo/hub/middleware.py b/kobo/hub/middleware.py
index bd4d423..68f7ecd 100644
--- a/kobo/hub/middleware.py
+++ b/kobo/hub/middleware.py
@@ -2,8 +2,6 @@

 from __future__ import absolute_import

-from django.utils.deprecation import MiddlewareMixin
-
 from .models import Worker


@@ -26,13 +24,21 @@ class LazyWorker(object):
         return request._cached_worker


-class WorkerMiddleware(MiddlewareMixin):
+class WorkerMiddleware():
     """Sets a request.worker.

     - Worker instance if username exists in database
     - None otherwise
     """

+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        self.process_request(request)
+        response = self.get_response(request)
+        return response
+
     def process_request(self, request):
         assert hasattr(request, "user"), "Worker middleware requires authentication middleware to be installed. Also make sure the database is set and writable."
         request.__class__.worker = LazyWorker()
```

This fixes the following traceback:
```python
  File "/usr/lib64/python3.13/threading.py", line 1041, in _bootstrap_inner
    self.run()
    ~~~~~~~~^^
  File "/usr/lib64/python3.13/threading.py", line 992, in run
    self._target(*self._args, **self._kwargs)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/django/utils/autoreload.py", line 64, in wrapper
    fn(*args, **kwargs)
    ~~^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/django/core/management/commands/runserver.py", line 143, in inner_run
    handler = self.get_handler(*args, **options)
  File "/usr/lib/python3.13/site-packages/django/contrib/staticfiles/management/commands/runserver.py", line 31, in get_handler
    handler = super().get_handler(*args, **options)
  File "/usr/lib/python3.13/site-packages/django/core/management/commands/runserver.py", line 73, in get_handler
    return get_internal_wsgi_application()
  File "/usr/lib/python3.13/site-packages/django/core/servers/basehttp.py", line 45, in get_internal_wsgi_application
    return get_wsgi_application()
  File "/usr/lib/python3.13/site-packages/django/core/wsgi.py", line 13, in get_wsgi_application
    return WSGIHandler()
  File "/usr/lib/python3.13/site-packages/django/core/handlers/wsgi.py", line 118, in __init__
    self.load_middleware()
    ~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/django/core/handlers/base.py", line 61, in load_middleware
    mw_instance = middleware(adapted_handler)
  File "/usr/lib/python3.13/site-packages/django/contrib/auth/middleware.py", line 121, in __init__
    super().__init__()
    ~~~~~~~~~~~~~~~~^^
TypeError: MiddlewareMixin.__init__() missing 1 required positional argument: 'get_response'
```

Note that this fix is only temporary, you will need to make additional changes to support Django 6.1: https://github.com/django/django/commit/1704c49a9b149b66b6a0e67abc8c95293bc35649

PR #321, opened 2025-07-11
packaging: use newer RPM macros for build/install on f42+

kdudka:
... because the old macros will likely stop working in Fedora 45

Resolves: https://bugzilla.redhat.com/2377365

### DIFF
--- a/osh.spec
+++ b/osh.spec
@@ -151,16 +151,15 @@ Conflicts: osh-worker-conf
 %description worker-conf-devel
 OpenScanHub worker devel configurations.
 
+%if 0%{?fedora} > 41 || 0%{?rhel} > 10
+%generate_buildrequires
+%pyproject_buildrequires
+%endif
+
 %prep
 %setup -q
 
 %build
-
-# Add -s to the shebang in osh/client/osh-cli:
-# https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_shebang_macros
-# TODO: Remove this when we migrate to newer Python packaging macros which
-# do this automatically.
-%py3_shebang_fix osh/client/osh-cli
 
 # collect static files from Django itself
 PYTHONPATH=. osh/hub/manage.py collectstatic --noinput
@@ -168,11 +167,24 @@ PYTHONPATH=. osh/hub/manage.py collectstatic --noinput
 # set path to python sitelib in the example httpd config
 sed 's|@PYTHON3_SITELIB@|%{python3_sitelib}|' osh/hub/osh-hub-httpd.conf.in > osh/hub/osh-hub-httpd.conf
 
+%if 0%{?fedora} > 41 || 0%{?rhel} > 10
+%pyproject_wheel
+%else
+# Add -s to the shebang in osh/client/osh-cli:
+# https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_shebang_macros
+%py3_shebang_fix osh/client/osh-cli
+
 %py3_build
+%endif
 
 
 %install
+%if 0%{?fedora} > 41 || 0%{?rhel} > 10
+%pyproject_install
+%else
 %py3_install
+%endif
+mv %{buildroot}{/usr,}/etc/
 
 # install the files collected by `manage.py collectstatic`
 cp -a {,%{buildroot}%{python3_sitelib}/}osh/hub/static/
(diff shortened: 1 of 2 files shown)


osh.spec
```diff
@@ -156,23 +156,32 @@ OpenScanHub worker devel configurations.
...
-%py3_shebang_fix osh/client/osh-cli
-
 # collect static files from Django itself
 PYTHONPATH=. osh/hub/manage.py collectstatic --noinput
 
 # set path to python sitelib in the example httpd config
 sed 's|@PYTHON3_SITELIB@|%{python3_sitelib}|' osh/hub/osh-hub-httpd.conf.in > osh/hub/osh-hub-httpd.conf
 
+%if 0%{?fedora} > 41
+%pyproject_wheel
```
hroncok, line 166: Using %pyproject_wheel without %pyproject_buildrequires is not supported. See https://lists.fedoraproject.org/archives/list/python-devel@lists.fedoraproject.org/thread/NMAI7IKK7ZK4VT2SRZCCUK4LLKJGZZ5D/

osh.spec
```diff
@@ -156,23 +156,32 @@ OpenScanHub worker devel configurations.
...
 %py3_build
+%endif
 
 
 %install
+%if 0%{?fedora} > 41
+%pyproject_install
+mv %{buildroot}%{python3_sitelib}/etc/ %{buildroot}/
+mv %{buildroot}%{python3_sitelib}/usr/{bin,share}/ %{buildroot}/usr/
+mv %{buildroot}%{python3_sitelib}/usr/lib/systemd/ %{buildroot}/usr/lib/
```
hroncok, line 181: This should not be necessary, see https://src.fedoraproject.org/rpms/pungi/pull-request/11#comment-268052

Instead, drop the `/usr` prefix from https://github.com/openscanhub/openscanhub/blob/8ba947af7fea949a453049d1dc720422aad872ed/setup.py#L21 https://github.com/openscanhub/openscanhub/blob/8ba947af7fea949a453049d1dc720422aad872ed/setup.py#L28 https://github.com/openscanhub/openscanhub/blob/8ba947af7fea949a453049d1dc720422aad872ed/setup.py#L31 https://github.com/openscanhub/openscanhub/blob/8ba947af7fea949a453049d1dc720422aad872ed/setup.py#L34

osh.spec
```diff
@@ -156,23 +156,32 @@ OpenScanHub worker devel configurations.
...
-# do this automatically.
-%py3_shebang_fix osh/client/osh-cli
-
 # collect static files from Django itself
 PYTHONPATH=. osh/hub/manage.py collectstatic --noinput
 
 # set path to python sitelib in the example httpd config
 sed 's|@PYTHON3_SITELIB@|%{python3_sitelib}|' osh/hub/osh-hub-httpd.conf.in > osh/hub/osh-hub-httpd.conf
 
+%if 0%{?fedora} > 41
```
hroncok, line 165: Do you also build this package on RHEL/EPEL? If so, please note that this will likely not work in RHEL 11+.

You would need:

```suggestion
%if 0%{?fedora} > 41 || 0%{?rhel} > 10
```

Or rhel > 8.

osh.spec
```diff
@@ -222,13 +231,17 @@ ln -s %{_sysconfdir}/osh/hub/settings_local.py %{buildroot}%{python3_sitelib}/os
...
 
 %files common
 %dir %{_sysconfdir}/osh
 %{python3_sitelib}/osh/common
 %{python3_sitelib}/osh/__init__.py*
 %{python3_sitelib}/osh/__pycache__
+%if 0%{?fedora} > 41
+%{python3_sitelib}/osh-*.dist-info
+%else
+%{python3_sitelib}/osh-*-py%{python3_version}.egg-info
```
hroncok, line 243: Why was this also moved to a different subpackage?

siteshwar:
The RPM build [failed](https://download.copr.fedorainfracloud.org/results/packit/openscanhub-openscanhub-321/fedora-42-x86_64/09267096-osh/builder-live.log.gz) with this error:

```
+ TMPDIR=/builddir/build/BUILD/osh-1.1.1.20250711.172741.g68785de.pr_321-build/osh-1.1.1.20250711.172741.g68785de.pr_321/.pyproject-builddir
+ /usr/bin/python3 -Bs /usr/lib/rpm/redhat/pyproject_wheel.py /builddir/build/BUILD/osh-1.1.1.20250711.172741.g68785de.pr_321-build/osh-1.1.1.20250711.172741.g68785de.pr_321/pyproject-wheeldir
/usr/bin/python3: No module named pip
```

osh.spec
kdudka: Fixed.

osh.spec
kdudka: When I drop the `/usr` prefix, the `osh-cli` binary is installed into `/usr/lib/python3.13/site-packages/bin/osh-cli` instead of `/usr/lib/python3.13/site-packages/usr/bin/osh-cli`, which is not an improvement really.

osh.spec
kdudka: Fixed.

osh.spec
hroncok: Sorry for the confusion, yes, I meant the `/usr/` prefix -- a.k.a. use paths relative to `/usr`.

This is not possible with `/etc` which is known. Wheels are packaged and than unpacked relative to prefix, and they cannot "leak" outside of it. Installing files to `/etc` from Python wheels is never going to work and we are expected to do this sort of thing on RPM level (either by moving the files or installing them manually or whatever).

osh.spec
kdudka: It might have been wrong previously (and nobody cared).  To be honest, I have no idea why we need to package these files.  My impression was that they should be installed whenever the related Python modules are installed, which means `osh-common` in this case.

Are you proposing to move them back to `osh-client`?  Would that improve anything?

PR #326, opened 2025-09-29
hub: optimize get_latest_scans()

sfowl:
This method performed one query to get all SystemReleases, then another for each SystemRelease to get all enabled scans, then again when .latest() is called. If we instead do this in a single query, returning all scans ordered by release tag and date submitted, then filter with python, then on pages like /osh/scan/packages which call this method there can be found and improvement to page load times of roughly 10x.

This is particularly noticeable in environments where there is noticeable network latency between hub instances and the backing database.

### DIFF
--- a/osh/hub/scan/models.py
+++ b/osh/hub/scan/models.py
@@ -284,22 +284,39 @@ of this packages scan")
     scans_number = property(calculateScanNumbers)
 
     def get_latest_scans(self):
-        srs = SystemRelease.objects.filter(active=True)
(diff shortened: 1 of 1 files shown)


sfowl:
We are carrying this patch internally as a workaround during a database migration. It is expected that once our migration is complete this patch will no longer be necessary, though may still offer a performance improvement.

kdudka (approved):
Looks good.  Thanks!

PR #336, opened 2026-02-09
hub: parse arches from cg_import tasks

sfowl:
Follow-up to dfea2f44, cg_import koji tasks do not have descendents with arch info, so instead let's parse arches directly from the cg_import task.

Also includes additional minor tweaks to reduce the number of getTag() koji calls.

PSSECAUT-1393


kdudka (approved):
Looks good.  Thanks!

PR #345, opened 2026-07-31
fix(hub): handle scan elapsed time over 24 hours

sfowl:
Extract parse_elapsed_time() into csmock_parser and use timedelta directly instead of strptime("%H:%M:%S"), which rejects hours > 23 and crashes result processing for long-running scans.

https://redhat.atlassian.net/browse/PSSECAUT-1603


kdudka (approved):
I did not test this but the changes look sane.

